PR #8, opened 2024-07-09
Optimize loading animation

HydrogenMacro:
The many divs in the loading checkerboard animation were removed and replaced by a background gradient.


Naviary2:
This is epic!! So much simpler and wayy less space! Can you also copy this over to the play.html and play.css code?

PR #10, opened 2024-07-09
Added a link to the source code in all of the footers

life-enjoyer:
A minor change to allow the users to easily find the source code from all pages


jermOSS:
In my opinion it would be better a bit better `target="_blank"`.

Naviary2:
This is great!

PR #14, opened 2024-07-09
Fix spelling of append from "apphend" across all files

pradyung:


Naviary2:
I have to go for now, I'll check this out later today!

Issue #16, opened 2024-07-09
Incorrect email regex


jermOSS:
This regex is incorrect.

https://github.com/Infinite-Chess/infinitechess.org/blob/52fb1674a87dd2eb0aaf9dc9e6814ce91dba7258/controllers/createaccountController.js#L217

Emails can contain more characters than just numbers, letters, dots and underscores.

Validating emails using regex is incorrect. Emails have very complex specification of how can they be composed. It is almost impossible to match all cases. This needs a better solution.

Nul-led:
You can send a cheap ICMP to the SMTP server after they have successfully bypassed the regex. Once they pass this we can actually send an email with the verification link.

Naviary2:
> You can send a cheap ICMP to the SMTP server after they have successfully bypassed the regex. Once they pass this we can actually send an email with the verification link.

This seems like a great idea to avoid sending emails to invalid or non-existent email accounts. Can you create a separate issue from this? I am going to be closing this one.

PR #19, opened 2024-07-09
Greater password security

Heinrich-XIAO:


Heinrich-XIAO:
This fixes #16

life-enjoyer:
It's great but the regex in [this post](https://stackoverflow.com/a/201378) is more complete I think. I've already made a PR with it: #18

Naviary2:
Screen Shot 2024-07-09 at 10 21 28 PM
I get this when a try to create a new account.

Btw the regex has been updated.

tsevasa:
Why is there a maximum password length of 30 characters? I would be in favor of increasing it to something larger, like 128, if there is no reason against it.

tsevasa:
I would also argue in favor of dropping the requirements for special characters, numbers, capitalized letters, etc. I feel like this might annoy people and infinite chess really doesn't need to be so restrictive, in my opinion.

Naviary2:
True, it does kind of seem excessive to have (such) strong password requirements, at least while this is still small, I can see it annoying people.

5jiji:
> Why is there a maximum password length of 30 characters? I would be in favor of increasing it to something larger, like 128, if there is no reason against it.

I'm wondering, why limit the length of the password at all?
Being that the password is hashed, why would we even enforce a length

tsevasa:
Ok, fair enough, you are right. Thinking about it again, I agree that there should be no maximum password length, it really makes no sense. Also, [this argument](https://stackoverflow.com/a/99724/19730773) makes sense:
```A maximum length specified on a password field should be read as a SECURITY WARNING. Any sensible, security conscious user must assume the worst and expect that this site is storing your password literally (i.e. not hashed).```

I would suggest to just enforce a minimum a password length of 8 characters and leave it at that.

PR #27, opened 2024-07-10
Adds a check for draw by insufficient material

NoobJsPerson:


Naviary2:
Hey! It doesn't work for me when it's KvKQ, but KvK works fine.

Naviary2:
This is really exciting! Insufficient material detection is desperately needed!

Naviary2:
This is closed in favor of the newer PR.

PR #36, opened 2024-07-10
IP Ratelimiter (Needs changing when restructre)

shacharrr:
Made the IP rate limiter work server-side.

The client sends his Ip address with his email and password in the request. and the server adds the client Ip as a key to the map with the properties:

`{ attempts: 1, timeout: 0 }`

every request new request increments the attempts count and when it reaches 3. the timeout incremented by 5 seconds. and after that the attempts reset to 1.

### DIFF
--- a/controllers/authController.js
+++ b/controllers/authController.js
@@ -102,7 +102,7 @@ function verifyBodyHasLoginFormData(req, res) {
         return false;
     }
 
-    if (typeof username !== "string" || typeof password !== "string") {
+    if (typeof username !== "string" || typeof password !== "string" || typeof ip !== "string") {
         console.log(`User ${username} sent a bad login request with either username or password not a string!`)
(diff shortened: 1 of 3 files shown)


controllers/authController.js
```diff
@@ -26,16 +28,35 @@ const refreshTokenExpirySecs = refreshTokenExpiryMillis / 1000;
 async function handleLogin(req, res) {
     if (!verifyBodyHasLoginFormData(req)) return; // If false, it will have already sent a response.
 
-    let { username, password } = req.body;
+    let { username, password, ip } = req.body;
```
Naviary2, line 31: If they are the ones to send their IP, they can modify the code to change it every single time to get around this. Get their ip from `middleware/IP/getClientIP()`

shacharrr: On it rn, didnt noticed this function sorry :)

Naviary2: All good!

Issue #4, opened 2024-07-02
Account Deletion


Naviary2:
There is no automated way yet for a user to delete their own account. Currently they have to email me, and I have to manually delete their account json data.

## Proposal ##
On your profile page, add a "Delete Account" button that when pressed will ask you to re-confirm your password, before logging you out and deleting your account.

Naviary2:
#62 fixed this

PR #37, opened 2024-07-10
Account deletion

jermOSS:
Fixes https://github.com/Infinite-Chess/infinitechess.org/issues/4


Naviary2:
This is awesome!!!

I think the button should be red to indicate it's a dangerous action. For example, here's one of github's dangerous buttons.

Screen Shot 2024-07-10 at 10 52 40 AM

And I think it should be at the bottom of the page, centered.

Screen Shot 2024-07-10 at 10 53 45 AM

And, more of a preference.. What if whenever a user deletes their account we log that info, to the console, and to a "deletedAccounts.txt" log?

Naviary2:
```js
logEvents(`User ${usernameLowercase} deleted their account.`, "deletedAccounts.txt", { print: true})
```

Naviary2:
Also, be sure to delete their role information. Inside controllers/roles.js, use `removeAllRoles()`!

jermOSS:
@Naviary2  What about now?

Naviary2:
What are some other opinions? I think this is completely ready now.

Screen Shot 2024-07-10 at 6 01 20 PM

Screen Shot 2024-07-10 at 6 01 46 PM

PR #39, opened 2024-07-10
Draw offers

this-is-not-available:
Not perfect, but it is a start


this-is-not-available:
Currently it only changes the dev version of the site.

this-is-not-available:
Also users can currently spam draw offers

this-is-not-available (commented):
Looks good 👍! I don't know the project completely, so I just try stuff to see what sticks.

this-is-not-available:
Spam prevention currently doesn't work because I forgot to update drawOfferMove in the client and the server after the checks

Naviary2:
This is great!!! Nice work. The functionality is there, I was able to successfully conclude the game by draw agreement! I'm going to make a list of things this still needs. Mostly polishing and bug fixes! Others are welcome to help fix these!

Screen Shot 2024-07-10 at 2 15 11 PM

Screen Shot 2024-07-10 at 2 16 11 PM

* The button should not be present in local games.
* Should not be present in online abortable games (less than 2 moves played)
* Should not be present after the game is over. Hide it when `gamefileutility.concludeGame()` is called.
* If you refresh the page while you have a draw offer it breaks it.
* We need some form of cap for how often you can send an offer. Maybe once every few moves?

And mostly just a preference, I think it should be less intrusive, maybe don't darken the whole screen, and place the popup in the lower middle of the board, or upper middle.

this-is-not-available:
Ok! I will strat work tomorrow. Starting with the limit and local game darkening.

this-is-not-available:
> And mostly just a preference, I think it should be less intrusive, maybe don't darken the whole screen, and place the popup in the lower middle of the board, or upper middle.

Because I'm lazy (and don't have much time), I decided that players shouldn't be able to make moves, while there is a draw offer.
It isn't server-sided yet or for the offerer, so that needs updating.
But this can change later on, if players or you want it to change.
For now, I'm sticking with this.

Naviary2:
That's what I mean about being less intrusive, I can only imagine how annoyed people will get if they can essentially *pause* their opponents game.

I'm thinking no black transparent backdrop, and a popup box either in the lower middle or the corner of the screen that does not interrupt your play.

Don't worry if you don't have time now, take your time. I'd rather this be polished than rushed.

this-is-not-available:
Ok

Naviary2:
This is a new design we came up with in the discord discussion 

Screen Shot 2024-07-11 at 7 35 48 AM copy

this-is-not-available:
> * The main menu and draw offer button need to be swapped so main menu is always on bottom.
> * The button should not be present in local games.
> * Should not be present in online abortable games (less than 2 moves played)
> * Should not be present after the game is over. Hide it when `gamefileutility.concludeGame()` is called.
> * If you refresh the page while you have a draw offer it breaks it.
> * We need some form of cap for how often you can send an offer. Maybe once every few moves?

These all have been addressed.

> And mostly just a preference, I think it should be less intrusive, maybe don't darken the whole screen, and place the popup in the lower middle of the board, or upper middle.

This is also addressed, but needs just a little more polish.

src/client/views/play.ejs
```diff
@@ -494,6 +495,13 @@
                         <div id="guardsB" class="promotepiececontainer hidden"><div class="promotepiece guardsB button"></div></div>
                     </div>
                 </div>
+                
+                <!-- Draw Offer UI -->
+                <div id="draw_offer_ui" class="draw_offer_ui footer hidden">
```
Naviary2, line 500: The draw offer UI needs to be moved to where my next comment says

this-is-not-available: Why

Naviary2: Actually it won't let me make a comment there so this is where it needs to go:

Screen Shot 2024-07-31 at 11 55 26 AM

Then, whenever the draw UI needs to reveal, in guigameinfo.js you have to use style.hideElement(element_whosturn) to hide the other words.

this-is-not-available: I tried to do that earlier, but it didn't work. Why do we need to move the UI up there? Shouldn't we be able to just hide the text but keep the UI where it is?

src/server/game/TypeDefinitions.js
```diff
@@ -126,6 +126,14 @@ function Game() {
      * AFK if they are currently AFK. */
     this.autoAFKResignTime = undefined;
 
+    /** Last move a draw was offered */
+    this.whiteDrawOfferMove = undefined
+    this.blackDrawOfferMove = undefined
+
+    /** Information about players offering a draw */
+    this.whiteDrawOffer = undefined
+    this.blackDrawOffer = undefined
```
Naviary2, line 135: First glace, it is unclear what types of values these are supposed to hold. It says "Information about...", are they strings? Are they booleans numbers? Please describe in the jsdoc

this-is-not-available: I'm unfamiliar with jsdoc, but I'm certain I can make something

src/client/views/play.ejs
Naviary2: What didn't work about it? The way you are doing it now, the UI is a separate object, overlayed on top of the already existing bar, it doesn't feel natural, it should be integrated into it, it should be a part of it.

src/server/game/gamemanager.js
```diff
@@ -1104,6 +1233,37 @@ const gamemanager = (function() {
...
+        const isOffering = (game.whiteDrawOffer === 'offered')
+        return isOffering
+    }
+
+    /**
+     * Returns *true* if the black in the provided game has a draw offer.
+     * @param {Game} game - The game
+     * @returns {boolean}
+     */
+    function hasBlackDrawOffer(game) {
```
Naviary2, line 1250: This could probably be merged with hasWhiteDrawOffer() to form like a new function hasColorDrawOffer(game, color)

src/client/scripts/game/misc/onlinegame.js
```diff
@@ -628,6 +669,8 @@ const onlinegame = (function(){
         }
 
         websocket.sendmessage('game', 'submitmove', data, true)
+        onlinegame.declineDraw() // decline draw offer, if any
```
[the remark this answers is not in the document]
this-is-not-available: yea changed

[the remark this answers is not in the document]
Naviary2: Great!

src/client/views/play.ejs
Naviary2: We do hide the text, only the draw offer UI or the whosturn element is visible at one time

PR #49, opened 2024-07-10
Fix esbuild removing global vars

5jiji:
This is the worst way to fix it, but it works, so why not?

(restructure branch)


5jiji:
This is useless now, as something has been found to be working.

PR #50, opened 2024-07-10
Per IP login rate limit (restructure branch)

shacharrr:
Made the IP rate limiter work server-side.

A map save every new IP address using the `middleware` function `getClientIP` with the properties: `{ attempts: 1, timeout: 0 }` 

every request increments the attempts count and when it reaches 3. the timeout incremented by 5 seconds. and after the timeout ends attempts reset back to 1.


Naviary2:
@shacharrr The restructure branch has been deleted! Please reopen this PR on the main branch!

Can you include more details about what it does exactly too? We already rate limit by IP, is this an upgrade? Is this rate limiting only the login route?

PR #52, opened 2024-07-10
Add checkForPieceCombinations for easier combination management and fix bugs in insufficientmaterial.js

NoobJsPerson:
I added the `checkForPieceCombinations` function to be able to check for combinations without the if chain.
If someone wants to add a combination they could modify the `pieceCombinationsForDrawCheckmate` and overall this makes it easier to add combinations or edit existing ones.

I also fixed some bugs with some piece combinations being skipped because there are 5 pieces on the board (I raised the piece count to skip the insufficient material check to 6).


Naviary2:
Conflicts with the main branch after merging the restructure branch with main. Please reopen.

PR #55, opened 2024-07-11
Getting rid of developmental play page

tsevasa:
Previous dev.html is now play.html. Now there is no more /play/devversion.


src/client/views/play.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
     <head>
-        <!-- Our in-document javascript is injected here before the server sends this html out -->
+        <!-- Our in-document javascript is injected here before we send out the file required to monitor the loading animation -->
```
Naviary2: The in document javascript **is** the file required to monitor the losing animation.

But it also is in charge of loading our audio.

src/client/views/play.html
```diff
@@ -10,25 +10,161 @@
...
+    <!-- The onload event listener to call main.start() is now within the htmlscript! -->
     <body>
         <header>
             <nav>
                 <a href="/">
                     <p>Home</p>
                 </a>
                 <a href="/play">
-                    <p class="currPage">Play</p>
+                    <p>Play</p>
```
Naviary2: This line can be reverted. Without it the "Play" link on the header isn't grey to indicate to you that is the page you're on.

src/client/views/play.html
```diff
@@ -221,6 +357,8 @@ <h2>Board Editing</h2>
                                         <option value="Abundance">Abundance</option>
                                         <option value="Amazon Chandelier">Amazon Chandelier</option>
                                         <option value="Containment">Containment</option>
+                                        <!-- <option value="Standarch - 3 Check">Standarch - 3 Check</option>
+                                        <option value="CoaIP - 3 Check">Coaip - 3 Check</option> -->
```
Naviary2, line 361: 👍

src/client/views/play.html
tsevasa: already did in my second commit

src/client/views/play.html
tsevasa: already reverted in my second commit

Naviary2: Oh my bad 😖 I thought I was viewing all changes but it must have been a single commit!

PR #61, opened 2024-07-11
Login attempts cooldown for each ip

shacharrr:
This one tries to tackle #2 

It is implemented as a map that contains the ip address of the one that try to join. every `maxLoginAttempts` failed attempts, the cooldown time increases by `loginCooldownIncrementorSec` seconds and the client can't login to any account in this time. when login succesfully. client ip is removed from the map.

structure of the map:
`loginAttemptData = {
    "127.0.0.1": { attempts: 1, cooldownTimeSec: 0 }
}`

this is a reopen of #50


Naviary2:
A thought I had is the keys of the loginAttemptData should be a combination of their IP **and** their username, so that if they have family members in the same household logging into their own account, they don't lock each other out.

For example, "shacharrr 127.0.0.1"

shacharrr:
That's a good idea, gonna add it asap

shacharrr:
Added now, every entry in `loginAttemptData` will look like this:

```js
{
  "ip": {
    "user1": { attempts: 1, cooldownTimeSec: 0 },
    "user2": { attempts: 1, cooldownTimeSec: 0 },
  }
}
```

src/server/controllers/authController.js
```diff
@@ -9,12 +9,17 @@ const bcrypt = require('bcrypt');
...
 
 const accessTokenExpiryMillis = 1000 * 60 * 15; // 15 minutes
 const refreshTokenExpiryMillis = 1000 * 60 * 60 * 24 * 5; // 5 days
 const accessTokenExpirySecs = accessTokenExpiryMillis / 1000;
 const refreshTokenExpirySecs = refreshTokenExpiryMillis / 1000;
 
+const maxLoginAttempts = 3;
+const loginCooldownIncrementorSec = 5;
+let loginAttemptData = {};
+
```
Naviary2, line 22: Can you add JSDoc comments for each of these variables? It's not exactly clear what they are for without having to find all their references.

src/server/controllers/authController.js
```diff
@@ -26,33 +31,64 @@ const refreshTokenExpirySecs = refreshTokenExpiryMillis / 1000;
...
 
+    
     let { username, password } = req.body;
     const usernameLowercase = username.toLowerCase();
     const usernameCaseSensitive = getUsernameCaseSensitive(usernameLowercase); // False if the member doesn't exist
     const hashedPassword = getHashedPassword(usernameLowercase);
 
     if (!usernameCaseSensitive || !hashedPassword) return res.status(401).json({ 'message': 'Username or password is incorrect'}); // Unauthorized, username not found
+    
+    clientIP = getClientIP(req);
```
Naviary2, line 42: Accidental global variable declaration.

src/server/controllers/authController.js
```diff
@@ -26,33 +31,64 @@ const refreshTokenExpirySecs = refreshTokenExpiryMillis / 1000;
...
+    if(!(clientIP in loginAttemptData)) {
+        loginAttemptData[clientIP] = {};
+    }
+
+    if(!(usernameLowercase in loginAttemptData[clientIP])) {
+        loginAttemptData[clientIP][usernameLowercase] = { attempts: 1, cooldownTimeSec: 0 };
```
Naviary2, line 48: Hmm.. a 2D array I think is a little too complex for this thing. I suggest doing something like this:

```js
const browserAgent = `${usernameLowercase}; ${clientIP}`;
if(!(browserAgent in loginAttemptData)) {
        loginAttemptData[browserAgent] = { attempts: 1, cooldownTimeSec: 0 };
    }
```

Naviary2:
A simpler method to rate limit them might be like this:

```js

const maxLoginAttemptsPerWindow = 5;
const windowSizeMillis = 1000 * 60; // 1 minute

// On password submitted...
const browserAgent = `${usernameLowercase}; ${clientIP}`;
if(!(browserAgent in loginAttemptData)) { // Verify their key exists first
        loginAttemptData[browserAgent] = 0;
}

loginAttemptData[browserAgent]++; // Increment their recent connection count
setTimeout(() => { // Set a timer to decrement their recent connection count after 1 minute
        loginAttemptData[browserAgent]--;
        if (loginAttemptData[browserAgent] === 0) delete loginAttemptData[browserAgent];
}, windowSizeMillis) // Decrement after 1 minute

if (loginAttemptData[browserAgent] >= maxLoginAttemptsPerWindow) {
        // Reject password attempt
        return;
}

// Accept password attempt...
// Test if it was correct...

```

jermOSS:
I think there is a bug in this code.

If these conditions are met the user is locked from any making further attempts forever:
1. Their amount attempts is >=`maxLoginAttempts`
2. Code crashes (or there is a restart) when code is waiting to decrease their attempts back to 1.
3. They make another wrong attempt.

This happens because now that their attempts are now at `maxLoginAttempts` from start.
This bypasses `if (loginAttemptData[clientIP][usernameLowercase].attempts > maxLoginAttempts) {` but since the password is incorrect it triggers `if (!match) {`, numbers of attempts is increased by one making it `maxLoginAttempts` + 1 bypassing `if(loginAttemptData[clientIP][usernameLowercase].attempts === maxLoginAttempts)` not triggering reset of attempts.

Next time this function is run by this user attempts are at `maxLoginAttempts`  + 1 triggering `if (loginAttemptData[clientIP][usernameLowercase].attempts > maxLoginAttempts) {` making them unable to reset amount of attempts.

jermOSS:
Using `setTimeout` is a poor design. It relies on server being able to hold state which is not a good idea. It also leads to having a lot of waiting `setTimeout`s which is not exactly good either.

It will be a lot better if you just store amount of attempts, length of cooldown and time of last attempt.

Also please include logs of failed attempts.

Emestrio:
The simplest solution would be to add the package "express-rate-limit" (https://express-rate-limit.mintlify.app/quickstart/usage).

The /login endpoint can have a specific rate limit restriction per IP. The rate limit can be a little higher than "normal" to take into account if multiple people on the same network login at the same time. But low enough to not waste resources doing password lookups when encountering spam. Keep in mind that normal traffic on this endpoint should be 1 or 2 attempts (per user) before being authenticated so the difference between this and artificial spam should be huge.

Naviary2:
> Using `setTimeout` is a poor design. It relies on server being able to hold state which is not a good idea. It also leads to having a lot of waiting `setTimeout`s which is not exactly good either.

Can you expand further on why this isn't a good idea? The server currently rate limits all requests in this manner. Is there some kind of cap on how many timers you can have at once?

this-is-not-available:
Wait, if we store the username of the request, can't the hacker send 10 billion characters long username and spam that 5 times a minute? That could potentially run the server out of memory and potentially deny service.

shacharrr:
On the username length, we should just enforce a maxUsernameLength or something that everything will abide by

shacharrr:
And after I read all of the comments here I think the best approach will be to do:

```js
{
    "ip": {
        "user1": { attempts: 0, cooldownTimeSec: 0, lastAttemptTime: <Date/Something else> },
        "user2": { attempts: 0, cooldownTimeSec: 0, lastAttemptTime: <Date/Something else> },
    }
}
```
I don't really see why to to this `ip username` stuff

shacharrr:
I added the changes you told me about, apart from the memory leak possibility with the username length as this should just get sorted on a global scale

Naviary2:
> I don't really see why to do this `browserAgent = "ip username"` stuff

I'm still in favor of doing this. While your method *does* work, 99.9% of the time there will not be someone in the same household logging into a separate account, so this IP object will almost always have 1 property, which is their username.

`loginAttemptData` should distinguish any single device, which can be done by a combination of their IP and username. `\`${usernameLowercase}${IP}\`` distinguishes a unique device, I don't see a reason to have it more complex with a double nested object. Maybe better to keep things simpler? Then you only have to check if 1 property exists instead of 2.

Naviary2:
> Wait, if we store the username of the request, can't the hacker send 10 billion characters long username and spam that 5 times a minute? That could potentially run the server out of memory and potentially deny service.

Right now it doesn't reach the limiter if the username is not found in the database, so that shouldn't be a problem.

shacharrr:
> > I don't really see why to do this `browserAgent = "ip username"` stuff
> 
> I'm still in favor of doing this. While your method _does_ work, 99.9% of the time there will not be someone in the same household logging into a separate account, so this IP object will almost always have 1 property, which is their username.
> 
> `loginAttemptData` should distinguish any single device, which can be done by a combination of their IP and username. `${usernameLowercase}${IP}` distinguishes a unique device, I don't see a reason to have it more complex with a double nested object. Maybe better to keep things simpler? Then you only have to check if 1 property exists instead of 2.

Im sorry if Im still understand it falsely, but my method is able to handle more than one username on each IP. so if there is one pc in a household that Is locked on `user1` another pc can log in properly to `user2`

shacharrr:
I can see your point about the complexity of this. and I can see what you mean.
I will change to the method you proposed as it really more simplictic

Naviary2:
> Im sorry if Im still understand it falsely, but my method is able to handle more than one username on each IP. so if there is one pc in a household that Is locked on `user1` another pc can log in properly to `user2`

The way I mentioned still allows more than one username per IP! If accounts "billy" and "joe" have ip "128.0.0.1", and they both try to log in at the same time with incorrect passwords, billy's key within `loginAttemptData` will be "billy128.0.0.1", and joe's key will be "joe128.0.0.1", very different!

Naviary2:
@shacharrr All right, I made a commit, basically cleaning stuff up and moving your logic to external functions so as to keep the main `handleLogin()` function small.

I also add an automatic timers that will delete browser agents from the login attempt data if they stop trying after some time. Then the login attempt data doesn't get cluttered up overtime when people give up trying to login!

Ready to merge! Thanks for your help!!! We can mark this issue as completed :) #2
No more risk of hackers brute force'ing your password!

Issue #20, opened 2024-07-09
Stricter password restrictions


Heinrich-XIAO:
Currently, a password of `123456` is allowed even though `password` is not. 

# Proposal
Impose more restrictions on the passwords.
For everyone that already has a password set as a "weak" password, they will receive popups, notifications, or alerts advising them to change their password.

life-enjoyer:
The passwords are hashed and salted so we can't know if someone already registered has a weak password.
But we could set more restrictions on newly created password such as having at least one special character, one uppercase letter and one number

Heinrich-XIAO:
That could happen on login though.

Naviary2:
Passwords are collected when logging in, so they can be checked for security at that time, they are neither saved, nor logged.

#19 Link to the pull request adding stronger password security requirements.

Naviary2:
At this time, we won't enforce stricter password requirements. There isn't enough at stake if your account's hacked, and the website is still relatively small, don't want to scare people off. In the future we can revisit this!

PR #67, opened 2024-07-11
Remove upper password length bound.

Heinrich-XIAO:


Heinrich-XIAO:
#20

tsevasa:
You corrected the if-check but the error message on that line still says "'Password must be between 6-30 characters long'".

src/server/controllers/createaccountController.js
```diff
@@ -251,7 +251,7 @@ const isValidEmail = function (string) {
 
 const doPasswordFormatChecks = function (password, res) {
     // First we check password length
-    if (password.length < 6 || password.length > 30) return res.status(400).json({ 'message': 'Password must be between 6-30 characters long'});
+    if (password.length < 6) return res.status(400).json({ 'message': 'Password must be between 6-30 characters long'});
```
Naviary2, line 254: (The message sent to the client also still says between 6-30)

Naviary2:
Is an unlimited password size going to allow people to internal-DDOS the server by sending extremely long passwords? This regex is run for every password submitted:

```js
const isValidPassword = function (string) {
    const regex = /^[a-zA-Z0-9!@#$%^&*\?]+$/;
    if (regex.test(string) === true) return true;
    return false;
}
```
To me this doesn't look complicated, I think this would be fast still. But who has more intuition for regexes?

Heinrich-XIAO:
Its O(n), so it can't be too bad.

src/client/scripts/createaccount.js
```diff
@@ -167,7 +167,7 @@ passwordInputElement.addEventListener('input', (event) => { // When password fie
     if (formatError) {
         passwordError.textContent = 'Password is in an incorrect format';
     } else if (lengthError) {
-        passwordError.textContent = 'Password must be between 6-30 characters long';
+        passwordError.textContent = 'Password must be more than 6 characters long';
```
Naviary2, line 170: Just a nitpick, but "atleast 6" 😆

Heinrich-XIAO:
> Is an unlimited password size going to allow people to internal-DDOS the server by sending extremely long passwords? This regex is run for every password submitted:
> 
> ```js
> const isValidPassword = function (string) {
>     const regex = /^[a-zA-Z0-9!@#$%^&*\?]+$/;
>     if (regex.test(string) === true) return true;
>     return false;
> }
> ```
> 
> To me this doesn't look complicated, I think this would be fast still. But who has more intuition for regexes?
I can't get it to take more than 0.01 seconds on a github codespace.

Naviary2:
Screen Shot 2024-07-11 at 2 54 06 PM

Also need to remove this attribute on the html element

Heinrich-XIAO:
Why was that originally 40 if the max is 30?

Naviary2:
That's a discrepancy!

Also I checked and the login.html has a max length of 40 as well, gotta remove that too

Idonotus:
truely unlimited passwords may lead to DoS attacks so have a high limit like 1024 of 256

Naviary2:
When I create an account with an extremely long password, when I login again, ANY password I enter that matches at least the first approximate 64 digits will log me in no matter if digits after those first 64 are incorrect.

Is this the bcrypt package that caps the password lengths? Is it even worth allowing unlimited password size if beyond 64 gives no more security?

Heinrich-XIAO:
How long is the "extremely long password"? 100? 200? 1000? 10000? or just more than 64?

Naviary2:
> How long is the "extremely long password"? 100? 200? 1000? 10000? or just more than 64?

Around 1000

5jiji:
> > How long is the "extremely long password"? 100? 200? 1000? 10000? or just more than 64?
> 
> Around 1000

That is extremely good, as my pasword manager limits me to make 128 chars long password (auto-generated) for some reason

Naviary2:
If we're going to cap it, we should probably do so at exactly bcrypt's limit.

Screen Shot 2024-07-11 at 8 32 09 PM

Naviary2:
> That is extremely good, as my pasword manager limits me to make 128 chars long password (auto-generated) for some reason

72 is not as good 😄 Which is the limit of bcrypt which we're using

Naviary2:
I suggest cap at 72. But if they have less than 6 characers we only say "Password must be 6+ characters", and if they try to enter more than 72 we say "Password must be not more than 72 characters." Most people don't make passwords that long so we don't need to tell them the limit unless they exceed it

5jiji:
> Bcrypt hashed password and secrets have a 72 characters limit.

Is it me, or i'm reading that 72 is the max characters for a **hashed** password, and not a \*true\* password?

Naviary2:
> Is it me, or i'm reading that 72 is the max characters for a **hashed** password, and not a *true* password?

All the hashes are exactly 60 characters you can check in members.json. I already tested out creating a long password and confirmed that any password over 72 characters essentially ignores what you put after that.

this-is-not-available:
I think part of the hash is metadata (hash algorithm, cost) so the hash part of the hash is less than that.

Heinrich-XIAO:
Does that mean that the limit should be 72 or that we need a different hashing library/algorithm?

this-is-not-available:
It means that more passwords will share the same hash, if encoded similarly. A prime example of the pigeonhole principle.

Naviary2:
Why should we search for an alternate package that allows infinite password length, when 99.9% of people don't make passwords that long, and, we'd still have to use bcrypt for testing passwords of users who created their account before the switch to the new package?

Heinrich-XIAO:
So, the maximum should just be 72?

Naviary2:
> So, the maximum should just be 72?

YesI think so!

Naviary2:
Ready to be merged!

PR #73, opened 2024-07-12
Setup translations framework

jermOSS:
Setup framework to allow contributors translate the app making Infinite Chess more accessible.
## TODO:
- [x] Install dependencies
- [x] Setup i18n with correct routing
- [x] Migrate existing text to translation files
- [x] Migrate existing HTML to EJS with i18n translations
- [x] Caching of templates
- [x] Add language switcher
- [x] Add documentation for translators


NoobJsPerson:
we need i18n or i18next not both. i18next seems to be most maintained so i guess it could be the best option

jermOSS:
I accidentally included the wrong package https://github.com/Infinite-Chess/infinitechess.org/pull/73/commits/a24d62e0e6d307f2f7c0562fbace8c99240a55e6 should include the right ones.

Naviary2:
Looks ready!!! Thanks @jermOSS @tsevasa @NoobJsPerson for your work on this! Soon Infinite Chess won't have any language barriers!

PR #81, opened 2024-07-14
Overhaul of file deployment in dev mode and production mode

tsevasa:
Please test this with DEV_BUILD turned on and off, respectively. It works perfectly fine for me.


tsevasa:
ToDos:
- Currently, the copy method copies over many unnecessary game scripts to dist that are never called by an HTML. Although it doesn't matter, we should only copy what is necessary.
- Revert bundling for development mode.

Naviary2:
Looking great! It's very fast for me too! 👏

Screen Shot 2024-07-14 at 12 00 42 AM

I fixed this error! The validation script is not directly part of the game code.

Basically what's left is (and I'm sort of re-iterating what you already put) is to
* In production, skip copying over the game code to dist (except for concatenating app.js).
* In dev, skip concatenating app.js. (We can create a new module "HTMLPreparator" that will inject the script tags we need into play.html's header, depending on if we're in dev or production.

tsevasa:
The TODO list is finished! In dev mode, now all scripts are just copied over to `dist`, while in prod mode, they are minified and concatenated into `app.js` where applicable.

Finally, the play.html page, now does not call any game scripts anymore. Instead, `HTMLScriptInjector.js` automatically recognizes the contents of `dist/scripts/game` and adds the appropriate script calls to it.

PR #82, opened 2024-07-14
Unverified Accounts Removal

shacharrr:
As stated in #1. This adds a function that is ran every 24hours/`intervalForRemovalOfOldUnverifiedAccounts`.
It checks for unverified accounts for more than `maxExistenceTimeForUnverifiedAccount`ms. and removes them from the DB.

Adds:
- `removeAccountByUsername(usernameLowercase, reason)`
- `removeOldUnverifiedMembers()` Running every interval and not meant to be used by other modules


jermOSS:
I have left few comments but other than that this code looks good to me.

Naviary2:
Ready to be merged!

PR #85, opened 2024-07-15
fixed missing return

RPink47:
fixed missing return on line 1224 of gamemanager.js in doesGameConclusionCheckOut so now correctly returns false if player cheated self mate


Naviary2:
Oh thanks! Funny thing this actually created a new bug where you can't submit again conclusion where to draw you one! I fixed it.

PR #86, opened 2024-07-15
replaced "nodemon" with "npx nodemon" in docs

tsevasa:
See relevant discussion [on discord](https://discord.com/channels/1114425729569017918/1257506171376504916).

This changed command allows users to run the server even when nodemon is not installed globally, see [npm nodemon](https://www.npmjs.com/package/nodemon). (`With a local installation, nodemon will not be available in your system path or you can't use it directly from the command line. Instead, the local installation of nodemon can be run by calling it from within an npm script (such as npm start) or using npx nodemon.`)

Apparently, several users already ran into this issue and solved it by installing nodemon globally as well, but that does not seem like a requirement one should make from users.

### DIFF
--- a/docs/NAVIGATING.md
+++ b/docs/NAVIGATING.md
@@ -2,7 +2,7 @@
 
 This guide gives you several pointers on how to navigate the project. The entire source code of the project is located in [src](../src/).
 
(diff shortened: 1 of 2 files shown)


Naviary2:
Looking great! Thanks!

Issue #68, opened 2024-07-11
Deleting Account doesn't require password


Heinrich-XIAO:
Deleting an account does not ask for the user's password. If an adversary were to gain access the the user's session token, but not the user's password, that adversary would be able to delete the user's account.

cycy98:
Maybe phone verification...
Because else we coulld brute force

tsevasa:
Or rather, I would propose e-mail + password verification:
When a user clicks "Delete account", they get sent an email with a temporary link. If they visit the link, they can input their password, to get the account deleted.
This would nicely complement #1 as unverified accounts will get deleted automatically anyways after some time.

Heinrich-XIAO:
> Maybe phone verification... Because else we coulld brute force

If #2 gets applied here, brute force won't be a big problem.

Naviary2:
Actually, any hacker will be able to send a request directly to the delete account controller, only requiring the users login info. If we are going to require your password to delete your account, that form data needs to be sent **with the request** to delete your account, and the remove account controller will have to be updated.

Naviary2:
Fixed by #89

PR #89, opened 2024-07-15
Add password requirement to account deletion

Heinrich-XIAO:
#68


Naviary2:
Cool!

src/server/controllers/removeAccountController.js
```diff
@@ -20,12 +21,21 @@ const intervalForRemovalOfOldUnverifiedAccountsMillis = 1000 * 60 * 60 * 24 * 1;
...
+async function removeAccount(req, res) {
     const usernameLowercase = req.params.member.toLowerCase();
+    const usernameCaseSensitive = getUsernameCaseSensitive(usernameLowercase); // False if the member doesn't exist
 
     // Check to make sure they're logged in
     if (req.user !== usernameLowercase) return res.status(403).json({'message' : 'Forbidden. This is not your account.'});
 
+		// Check that the password is correct
+		const hashedPassword = getHashedPassword(usernameLowercase);
+		const match = await bcrypt.compare(req.body.password, hashedPassword);
```
Naviary2, line 33: Use the new `testPasswordForRequest()` method I added in authController instead of using repeat logic here.

src/client/views/member.html
```diff
@@ -53,7 +53,7 @@ <h6 class="center">Account Info</h6>
                         </section>
                     </div>
                     <div style="flex-grow: 0;">
-                        <button id="delete-account" class="hidden action-button" onclick="removeAccount()">Delete account</button>
+                        <button id="delete-account" class="hidden action-button" onclick="removeAccount(true)">Delete account</button>
```
Naviary2, line 56: Instead of doing this, add an event listener on this element in the javascript that listens for the 'onclick' event. This way all javascript stays in our javascript file.

Heinrich-XIAO:
Should I change the alerts and prompts to the new [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) elements?

src/client/scripts/member.js
```diff
@@ -125,28 +125,32 @@ function showAccountInfo() {
...
+async function removeAccount(confirmation) {
+    if (!confirmation || confirm("Are you sure you want to delete your account? This CANNOT be undone! Click OK to enter your password.")) {
+        const password = prompt("Enter your password to PERMANENTLY delete your account: ");
         const config = { // Send with our access token
             method: 'DELETE',
             headers: {
                 'Content-Type': 'application/json',
                 'Authorization': `Bearer ${token}`
             },
+			body: JSON.stringify({ password }),
```
Naviary2, line 137: Add the `username` property in the body!

Heinrich-XIAO: But that's redundant as the username is determined by the Bearer token.

Naviary2: Ah you beat me! I deleted my comment and was about to respond to the other one.

src/server/controllers/removeAccountController.js
Naviary2: Hmm it may be redundant. You can set req.body.username on the server side, that's fine!

Heinrich-XIAO:
So, should I add username? Or do I just refactor `testRequestForPassword`

Naviary2:
> Should I change the alerts and prompts to the new [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) elements?

I'm kind of unfamiliar with those. Do you have screenshots? Probably eventually we will want to no longer use pop-up confirmations like this, because it's not as nice as an in-document form, but this is a great start!

Heinrich-XIAO:
This is without CSS. 
image

Heinrich-XIAO:
I haven't really worked with it either.

Heinrich-XIAO:
image

Naviary2:
> This is without CSS. image

Hmm, I don't know how I feel.. looks works than prompt boxes imo.

Heinrich-XIAO:
> > This is without CSS. image
> 
> Hmm, I don't know how I feel.. looks works than prompt boxes imo.

At least you can change it with CSS.

Heinrich-XIAO:
Plus its consident across browsers.

Naviary2:
> At least you can change it with CSS.

At that stage, we may as well create our own form on the page

Heinrich-XIAO:
Dialog is like a div, but is a dialog. You can put a form in it.

Heinrich-XIAO:
> > At least you can change it with CSS.
> 
> At that stage, we may as well create our own form on the page

That's what a dialog is for, but with some features bulit in.

Naviary2:
Hmm, do we really need the autoRespond parameter? Figure the testPasswordForRequest() method would be able to tell the client why it failed, whether it was incorrect, or if it was rate limited. Can we just display that error in another confirmation box before canceling the account removal?

src/client/scripts/member.js
```diff
@@ -110,6 +110,7 @@ function loadMemberData (loggedInAs) {
             revealElement(element_showAccountInfo)
             // Display remove button
             revealElement(element_deleteAccount)
+						element_deleteAccount.addEventListener("click", () => removeAccount(true));
```
Naviary2, line 113: 🔥

Heinrich-XIAO:
Does it really have to say `Username or password is incorrect` if anyone can just check if the username exists?

Naviary2:
True, but I think it's just a little more secure.

I imagine if you enter an incorrect password while trying to delete your account, the client would display the error message from the server in a confirm box, then it would reopen the enter password prompt.

Heinrich-XIAO:
> True, but I think it's just a little more secure.

How is it more secure? Because if it says the password is right, the username is right, but they can check that with the member page.

Naviary2:
> How is it more secure? Because if it says the password is right, the username is right, but they can check that with the member page.

I suppose it doesn't matter too much.

Ready to be merged!

PR #90, opened 2024-07-16
Added autofill compatibility to login

kingmancho:
Added a check after page load to see if login inputs were auto filled and updates the submit button accordingly. Before, the login button was grayed out but still clickable.

Also cleans up some code.


Naviary2:
Looks good!

PR #97, opened 2024-07-17
Premoves

BF5258:
Adding the option to make premoves.

## To Do:
- [x] Add premoves
- [x] Display the premove on the player's board
- [ ] Highlight the premoved piece in a different color
- [x] Allow multiple premoves to be made
- [x] Allow the same piece to be premoved multiple times
- [x] Display premove options when a piece is selected
- [x] Allow premoves to be canceled
- [ ] Fix bugs (see below)
- [ ] Add an "Enable premoves" option to settings
(Has anyone written a settings page?)
- [ ] Sound effect when premoving
- [ ] Polish / Clean up

### Current Bugs:
- [x] When a piece premoves out of a square it momentarily leaves behind a ghost piece that can still be selected.
- [ ] When the king premove castles the rook's position doesn't visually change.
- [ ] When a capture has been premoved the capturing piece sometimes renders underneath the piece captured.
- [ ] After a promotion premove has been made, the piece still renders as and displays the moves of a pawn.
- [x] "Special moves" such as pawn captures, double push, and castling sometimes don't display when they should.


BF5258:
Possible premoves are now shown. Except pawns:
En passant bug

src/client/scripts/game/misc/onlinegame.js
```diff
@@ -370,6 +373,47 @@ const onlinegame = (function(){
...
         flashTabNameYOUR_MOVE(true);
         scheduleMoveSound_timeoutID();
+
+        handlePremove();
+    }
+    /**
+     * Submits the next premove to the server if it is legal.
+     * Otherwise deletes the queue.
+     */
+    function handlePremove() {
```
Naviary2, line 383: This is weird to put this function in onlinegame.js. Let's create a new script called `premove.js` and put all the premove logic in there to keep it separated from the rest of the code!

Naviary2:
Looking good!

Screen Shot 2024-07-18 at 1 10 53 AM

Inside options.themes, lets add an option for the premove color, to separate them from normal moves!

src/client/views/play.html
```diff
@@ -157,6 +157,8 @@
 			onload="(() => { htmlscript.removeOnerror.call(this); })()"></script>
 		<script defer src="/scripts/game/chess/insufficientmaterial.js" onerror="htmlscript.callback_LoadingError(event)"
 			onload="(() => { htmlscript.removeOnerror.call(this); })()"></script>
+        <script defer src="/scripts/game/chess/premove.js" onerror="htmlscript.callback_LoadingError(event)"
```
Naviary2, line 160: This addition of the script into play.html is actually **totally automated** now!

Screen Shot 2024-07-19 at 1 56 53 AM

I just realized your fork is 143 commits behind the main branch! There are a lot of new changes you haven't seen yet! There should be a "sync" button on your fork page on github! You may have to resolve some conflicts if the main branch's code conflicts with your own.

Naviary2:
Is the client going to wait to submit premoves until it's your turn again, or is it going to submit instantly and the server keep track of that stuff? Cause I imagine the client would wait, it would be extremely difficult to get the server to keep track of premoves.

BF5258:
This doesn't look right:
image

BF5258:
I think I broke Rook:
image

Naviary2:
I just created a PR on this to merge the changes on the main branch to here so that it is up to date!

Naviary2:
> Because the server did not have code to validate moves at the time of writing, **all premoves are client based**.

Alright yes, I like this!

src/client/scripts/game/chess/game.js
```diff
@@ -82,6 +82,7 @@ const game = (function(){
         if (input.isKeyDown('tab')) guipause.callback_TogglePointers();
         if (input.isKeyDown('r')) piecesmodel.regenModel(game.getGamefile(), options.getPieceRegenColorArgs(), true);
         if (input.isKeyDown('n')) options.toggleNavigationBar();
+        if (input.isKeyDown('c')) premove.clearPremoves();
```
Naviary2, line 85: Should we stick to right-click cancelling all premoves?

Screen Shot 2024-07-22 at 12 16 06 PM

selection.js line 72

Naviary2:
I created a PR that allows you to click opponent pieces to view their legal moves. This is very close to utilizing premoves now because we also want the ability to click our own pieces when it's NOT our turn to view our legal moves... which at that time we should activate premoving!

src/client/scripts/game/chess/premove.js
```diff
@@ -0,0 +1,248 @@
...
+        highlightedSquares = [];
+    }
+
+    /** Displays premoved pieces in their new positions. */
+    function showPremoves() {
+        if(!premoves) return;
+        let gamefile = game.getGamefile();
+        for (let movedPiece of Object.values(movedPieces)) {
+            if(movedPiece.premovedCoords)
+                piecesmodel.movebufferdata(gamefile, movedPiece, movedPiece.premovedCoords);
```
Naviary2, line 174: Hmm. This seems a bit weird for premove.js to modify the mesh data of the gamefile. It may be better to just use `movepiece.makeMove()` instead. That handles everything for you from modifying the mesh data to animating the piece to sound effects to updating the organized lines. I imagine not using would be the source of many bugs.

Here's how it would work. When we make a premove, we call `makeMove()`. When we receive our opponent's move, at that time we use`movepiece.rewindGameToIndex()` to undo all our premoves, then we make our opponents move, then we check if the first premove in our list is legal, if so we make it, and re-do the the remainder of the premoves.

tsevasa:
Oh no, why was this deleted?

Naviary2:
> Oh no, why was this deleted and closed?

I don't know... it doesn't look like he's in the discord anymore either.

src/client/scripts/game/chess/premove.js
BF5258: Thank you. This has now been implemented.

BF5258:
Currently the check highlight is shown every second premove. This looks a bit weird. Should it be shown every move or not at all? Does updateInCheck need to be called for premoves?

https://github.com/user-attachments/assets/b1f62780-60eb-4505-ad42-6eece797f7f2

Naviary2:
Hmm. This would be where I'd want to study the behavior of premoving on the other chess websites. How do they render check highlights when premoving? Do they only render the check highlight of the check before any premoves are applied? Or the check after the premoves are applied?

If they render the check highlight after applying the premoves, I'm leaning towards modifying the function that checks if the current position is in check, to check if the opposite color of the piece that moved is in check, rather than checking if the opposite color of the color that *should* have moved is in check, as premoving doesn't care if you or your opponent *should* have moved that turn.

BF5258:
> This would be where I'd want to study the behavior of premoving on the other chess websites.

Parity with other websites is a good idea, but this situation seems to be exclusive to infinitechess.org. Chess.com never highlights check. Lichess.org doesn't move the piece until it is your turn. I am unaware of any websites that highlight check as well as allowing multiple premoves.
> ... check if the opposite color of the piece that moved is in check....

This would work but could be more specific. What will "opposite" mean in four player? The next player or all of the other players? It's also useful to see the highlight on your own pieces. If a premove has been made into check, it is unlikely to be legal unless the opponent has a reason to move the attacking piece away.

>Oh no, why was this deleted and closed?

I don't remember why I deleted it. I was stressed by an influx of school assignments but deleting the repository was a bit drastic. It's back now: https://github.com/BF5258/infinitechess.org/tree/premoves

Naviary2:
Okay yeah, other websites don't highlight checks with premoves, let's adopt that.

> What will "opposite" mean in four player? The next player or all of the other players?

I feel like it would be all other players. But, that's redundant now if we're not going to be rendering check highlights. We just need out check-finder function to early-exit if there are any premoves present.

PR #115, opened 2024-07-21
Polish Translation

cnnn666v2:
I've translated some of the texts into polish, will continue to translate more throughout the days


Naviary2:
Let me know when it's ready! Could you mark this PR as a draft until then?

tsevasa:
Note that [en-US.toml](https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/en-US.toml) just got updated to version number 5. Please see [changes.json](https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/changes.json) for the changelist, it is quite extensive. I recommend updating your current translation as soon as possible to reflect these changes. I am very sorry for the inconvience, we needed to still add some missing localizations to the website but no more such changes are foreseen for now.

cnnn666v2:
> Let me know when it's ready! Could you mark this PR as a draft until then?

Done!



> Note that [en-US.toml](https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/en-US.toml) just got updated to version number 5. Please see [changes.json](https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/changes.json) for the changelist, it is quite extensive. I recommend updating your current translation as soon as possible to reflect these changes. I am very sorry for the inconvience, we needed to still add some missing localizations to the website but no more such changes are foreseen for now.

Sure mate :D

cnnn666v2:
So upon translating some texts, I noticed that in some areas polish translation will be a little "broken". The thing with polish language is that we have different endings for words depending on the words before it. It's most visible with time, so in english you would say "1 second", "2 seconds", "5 seconds"; but in polish it would be "1 sekun**da**", "2 sekun**dy**", "5 sekun**d**".
My question is if it would be possible to add another property for that case? Numbers ending with a 2 - 4 would have the **dy** ending, numbers 0, 5 - 9 would end with a **d**. And meanwhile a 1 on its own ends with a **da**, when you put it at the end of any other number (101,  1021,  51, etc.) would end with a **d**.
Sorry if it's too confusing but I tried my best to explain it, if it's not possible to do so then it's alright.

tsevasa:
Oh, good point. Probably, also other languages will have this issue of multiple pluralities... In principle, it's doable to adapt the code to accommodate that. For example, lines like this could be adapted to check more cases than just "equals 1" or "not equals 1":
```js
const seconds_plurality = login_cooldown == 1 ? translations["ws-second"] : translations["ws-seconds"];
result_message += ` ${login_cooldown} ${seconds_plurality}.`
```
and then we would have a table of all possible pluralities for all possible distinct values of `login_cooldown`, in this case.

I wonder if that is worth it though since this table could get very large eventually. Does there exist a commonly used abbreviation for "sekund/da/dy" in Polish? (like "sec" in English for example) Maybe using abbreviations would make our lives a lot easier.

cnnn666v2:
Hm, I think that in polish you could use both "s" and "sek". Once I get back to it I should change the texts to this abbreviation, thx

tsevasa:
Great, in that case "sek" and "min" would be ideal!

Naviary2:
I'm going to merge this now so that others can contribute to the Polish translation as well! @cnnn666v2 you can continue to work on it and reopen new pull requests when you return!

cnnn666v2:
Oh yeah, nice! I kind of lost motivation again for everything github related but I'll try to get back to work on this translation. One thing I have to ask you though, what does the "Rated" mean in the game browsing menu? Like, what does it mean that the match is or isn't rated?

Naviary3:
> Oh yeah, nice! I kind of lost motivation again for everything github related but I'll try to get back to work on this translation. One thing I have to ask you though, what does the "Rated" mean in the game browsing menu? Like, what does it mean that the match is or isn't rated?

@Apsurt has finished the remainder of it, you can view it [here](https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/pl-PL.toml)!

A chess match can either be "Rated", or "Casual". If it's rated, then ELO points are at stake, and the winner will take some points from the loser. Casual matches are more chill.

PR #116, opened 2024-07-21
Added many missing localizations

tsevasa:


src/server/game/invitesmanager.js
```diff
@@ -273,7 +273,8 @@ function areUnderMaintenance(ws) {
     gamemanager.printActiveGameCount();
     let message = allowinvites.message;
     const timeUntilRestart = getMinutesUntilRestart();
-    if (timeUntilRestart) message += ` Minutes: ${timeUntilRestart}`
+    // TODO: send the timeUntilRestart with the message somehow
+    // if (timeUntilRestart) message += ` Minutes: ${timeUntilRestart}`
```
Naviary2, line 277: I think it's better to include the time until the server restarts with this message, even though it may not translated yet, they'd have no way of knowing how soon.

Naviary2:
Thanks for updating the variant scripts server side too!

tsevasa:
> Thanks for updating the variant scripts server side too!

God bless the Project Search functionality of VSCode, I would be lost without it :D It makes it so easy to find all occurences of some phrase in the repo.

src/server/controllers/createaccountController.js
```diff
@@ -190,9 +190,9 @@ function checkUsernameAvailable(req, res) {
     let allowed = true;
     let reason = '';
 
-    if (doesMemberExist(usernameLowercase)) { allowed = false; reason = 'That username is taken'; }
-    if (checkProfanity(usernameLowercase)) { allowed = false; reason = 'That username contains a word that is not allowed'; }
-    if (reservedUsernames.includes(usernameLowercase)) { allowed = false; reason = 'That username is taken'; } // Code for reserved (but the users don't know that!)
+    if (doesMemberExist(usernameLowercase)) { allowed = false; reason = "ws-username_taken"; }
+    if (checkProfanity(usernameLowercase)) { allowed = false; reason = "ws-username_bad_word"; }
+    if (reservedUsernames.includes(usernameLowercase)) { allowed = false; reason = "ws-username_taken"; } // Code for reserved (but the users don't know that!)
```
Naviary2, line 195: Maybe we should give this one it's own code, in case in the future we do want to tell them if it's reserved, vs taken. I mean otherwise how would these big reserved names know they could claim their name?

src/server/game/gamemanager.js
```diff
@@ -770,7 +770,7 @@ const gamemanager = (function() {
 
         if (isGameOver(game)) { // Resync them to the game because they did not see the game conclusion.
             console.error("Player tried to resign game when the game is already over!")
-            ws.metadata.sendmessage(ws, 'general', 'notify', "Can't resign game, it's already over.")
+            ws.metadata.sendmessage(ws, 'general', 'printerror', "Can't resign game, it's already over.")
```
Naviary2, line 773: I'm in favor of this getting it's own code for the 'notify' action, because otherwise the player will have thought they resigned, not knowing it ended a split second before then.

src/server/game/variant1.js
```diff
@@ -217,29 +217,31 @@ const variant1 = (function() {
...
                 initKnightline(gamefile, { Variant, Date })
                 break;
-            case "Knighted Chess":
+            case "Knighted_Chess":
                 initKnightedChess(gamefile, { Variant, Date })
                 break;
             // Showcasings...
             case "Omega": // Joel & Cory's version
-                variantomega1.initOmega(gamefile, { Variant, Date });
+                variantomega.initOmega(gamefile, { Variant, Date });
```
Naviary2, line 225: Whoops! You accidentally referenced the wrong script! Should be `variantomega1`. This is hard to catch. For a fast way to convert them all, rename the "variantomega1" import at the top of the module to "variantomega", thenafter Right Click > Rename symbol, and rename back to "variantomega1". This will change all instances of "variantomega" in the module to "variantomega1"!

src/server/game/variant1.js
```diff
@@ -439,7 +443,7 @@ const variant1 = (function() {
      * @param {Object} [position] - The starting position of the game, organized by key `{ '1,2': 'queensB' }`, if it's already known. If not provided, it will be calculated.
      * @returns {Object} The gamerules object for the variant.
      */
-    function getGameRulesOfVariant({ Variant, Date = math1.getUTCDateTime() }, position) {
+    function getGameRulesOfVariant({ Variant, Date = math.getUTCDateTime() }, position) {
```
Naviary2, line 446: Same as above! Follow the same steps to rename all instances of `math` to `math1`

src/server/controllers/createaccountController.js
tsevasa: Hm, I thought I remembered a conversation where the current version was settled upon and I didn't change any text when localizing, but it's fine by me anyways to change it. I suppose there is a nonzero chance and we should add new localizations sooner rather than later.

src/server/game/variant1.js
tsevasa: Oh, my bad! Apparently this was wrong in a few more places in that script that I did not touch.

src/server/game/variant1.js
tsevasa: Done!

src/server/controllers/createaccountController.js
tsevasa: Done!

tsevasa:
Thank you for the feedback! I implemented it and marked the conversations above as "resolved".

src/server/game/invitesmanager.js
tsevasa: Done!

Naviary2:
Thanks for the changes! Looks good!

PR #124, opened 2024-07-22
Optimise images

jermOSS:
Converts images into more efficient formats like AVIF or WebP using sharp library.


Naviary2:
Looks good now! Thanks!!! All image assets are now so much smaller! 🎉

PR #126, opened 2024-07-23
Add localised last seen messages

NoobJsPerson:


src/server/controllers/memberController.js
```diff
@@ -6,6 +6,7 @@
  * And we resend requests account verification emails.
  */
 
+const locale = require('date-fns/locale')
```
Naviary2, line 9: Can you merge this with the line beneath? Those imorts are from the same package.

NoobJsPerson: deconstructing doesn't work i have to work the / syntax

NoobJsPerson: probably because the package doesn't export local so its the only way to access locale

NoobJsPerson: oh its because its a sub-module thats why

PR #127, opened 2024-07-24
fix redundancy

pihedron:
# Redundancy Fix

Nothing has changed much in terms of functionality. Most of the edits are to do with formatting, but I essentially created a layouts folder with a `default.ejs` file that serves as a *skeleton* for most of the pages.  I wrote some code in [`src/server/config/setupTranslations.js`](https://github.com/pihedron/infinitechess.org/blob/5b1f440/src/server/config/setupTranslations.js) that passes some additional arguments through the `ejs.render` function so that HTML from other files can be included into the skeleton. This means we don't have to repeat the `head` and `body` tags for each page and we can dynamically highlight the navigation bar.

## Details

- no changes to dependencies
- CSS is **not injected** but linked
- retains the same functionality as translation

> I edited this comment to include more information.


pihedron:
I will work on the redesign next but I think it's vital to get this done first.

Naviary2:
Hey! Can you describe in detail how it works? Does it use another package as a dependency? Does it inject common.css into every html's css file before serving? Does it add common.css as a source for each html?

pihedron:
# Redundancy Fix

Nothing has changed much in terms of functionality. Most of the edits are to do with formatting, but I essentially created a layouts folder with a `default.ejs` file that serves as a *skeleton* for most of the pages.  I wrote some code in [`src/server/config/setupTranslations.js`](https://github.com/pihedron/infinitechess.org/blob/5b1f440/src/server/config/setupTranslations.js) that passes some additional arguments through the `ejs.render` function so that HTML from other files can be included into the skeleton. This means we don't have to repeat the `head` and `body` tags for each page and we can dynamically highlight the navigation bar.

## Details

- no changes to dependencies
- CSS is **not injected** but linked
- retains the same functionality as translation

Naviary2:
SWEET!!! Thanks! I'll check it out soon! Hey uh.... I merged another PR which created a lot of conflicts, could you resolve those? 😅

pihedron:
I am going to close this and open a new one to see if I can fix these merge conflicts in a better way.

PR #135, opened 2024-07-28
fix spelling and landing page

pihedron:
# Redesign Process

I fixed some spelling mistakes present in different parts of the code. I would recommend installing something like **prettier** so our JavaScript is consistent across people because I keep seeing double quotes and single quotes in the same place. The TOML file had some spelling mistakes as well. Besides that, I centered the splash text on the landing page and brought the About & Patreon section back.

## Details

- [ ] dependency changes
- [ ] restructuring
- [ ] game changes
- [x] template changes
- [x] script changes
- [x] style changes
- [x] translation changes
- [x] formatting


src/client/views/components/head.ejs
```diff
@@ -6,7 +6,7 @@
 <meta charset="utf-8" />
 <meta http-equiv="x-ua-compatible" content="ie=edge" />
 <meta name="viewport" content="width=device-width, initial-scale=1" />
-<link rel="icon" type="image" href="/img/favicon.png">
+<link rel="icon" type="image" href="/img/logo.png">
```
Naviary2, line 9: I don't think it's standard to have the tab icon named logo..

src/client/scripts/memberHeader.js
```diff
@@ -144,4 +143,17 @@ const validation = (function(){
         areWeLoggedIn,
     })
 
+})();
+
+const pihedron = (function () {
```
Naviary2, line 148: "Pihedron" doesn't explain what this does. Can you just move its logic into a function called, maybe setActiveHeaderLink, or something, and put that function inside the above IIFE?

src/client/views/layouts/play.ejs
```diff
@@ -27,4 +27,5 @@
         </main>
         <%- include(`${viewsfolder}/components/footer`, {t:t, languages:languages, language:language}) %>
     </body>
+    <script src="/scripts/memberHeader.js"></script>
```
Naviary2, line 30: Can this be moved to in the head by the other scripts? It feels of of place to be at the bottom of the document

src/client/views/news.ejs
```diff
@@ -1,6 +1,11 @@
 <div id="content">                  
     <h1 class="center"><%=t('news.title')%></h1>
 
+    <p class="date"><%=t('news.july28-2024.date')%></p>
```
Naviary2, line 4: I'd prefer if we left news announcements to me.. Also people don't see new news posts until I actually push an update to the website, so it throws off dates...

Naviary2: At the time of a website update, I will provide a changelog.

src/client/views/play.ejs
```diff
@@ -72,7 +72,7 @@
             <p>
                 <%=t('play.guide.rules_paragraphs.0')%>
             </p>
-            <img data-src="/img/game/guide/promotionlines.png" class="img-promotionlines">
+            <img src="/img/game/guide/promotionlines.png" class="img-promotionlines">
```
Naviary2, line 75: These images are actually intended to have data-src attributes instead of src, this is so that we can dynamically load the guide images only when they open the guide, instead of in page load. This saves bandwidth and improves loading time.

translation/fr-FR.toml
```diff
@@ -28,7 +28,7 @@ js-enter_password = "Entrez votre mot de passe pour supprimer votre compte DÉFI
 [index]
 secondary_title = "Le site officiel pour jouer en direct !"
 what_is_it_title = "Qu'est ce que c'est ?"
-what_is_it_pargaraphs = [
+what_is_it_paragraphs = [
```
Naviary2, line 31: Changes to the TOML files need to be specified in translations/ changes.json. But it's okay if you didn't know that

src/client/views/components/head.ejs
Naviary2: A favicon is a browser icon, short for favorite icon. Many browsers will auto-request a favicon image if it exists because it's so commonly given that name. If it's named differently it's not immediately clear that it is a browser icon, there might even be extra requests, with the browser fetching logo.png *and* seeing if we have a "favicon" as well.

src/client/views/play.ejs
Naviary2: If you made this change to fix the images not loading (if they weren't loading), then there's another bug that's causing them to not load when clicking the Guide.

src/client/views/components/head.ejs
Naviary2: We now have a new server icon!

PR #136, opened 2024-07-28
Added legal move highlight outside of a player's turn

tsevasa:
In online games, when it's not a player's turn, clicking their own pieces shows what their legal moves would be.

### DIFF
--- a/src/client/scripts/game/chess/selection.js
+++ b/src/client/scripts/game/chess/selection.js
@@ -14,6 +14,9 @@ const selection = (function() {
     /** Whether or not the piece selected belongs to the opponent.
      * If so, it's legal moves are rendered a different color, and you aren't allowed to move it.  */
     let isOpponentPiece = false;
+    /** Whether or not the piece selected activated premove mode.
+     * This happens when we select our own pieces, in online games, when it's not our turn. */
+    let isPremove = false;
 
     /** The tile the mouse is hovering over, OR the tile we just performed a simulated click over: `[x,y]` */
     let hoverSquare; // Current square mouse is hovering over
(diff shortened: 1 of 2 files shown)


src/client/scripts/game/chess/selection.js
```diff
@@ -124,6 +124,10 @@ const selection = (function() {
             return;
         }
 
+        // Don't allow players to move a piece of a color not matching the current player's turn
+        const selectedPieceColor = math.getPieceColorFromType(pieceSelected.type);
+        if (selectedPieceColor !== gamefile.whosTurn) return;
```
Naviary2, line 129: Hey! Are these few lines needed! I can't find a reason, as `hoverSquareLegal` is always false if we've selected a friendly piece when it's not our turn.

tsevasa: You are right that they are not really needed right now, because I added `selectionColorAgreesWithMoveTurn` to line 253, setting `hoverSquareLegal` to false in that case. My thinking was that we will want to remove that check in line 253 in the future again when we implement premoves because we will want to show the ghost pieces for the premoves, so `hoverSquareLegal` will need to be true, even when it is not our turn. In that case, we will need the currently superfluous lines again to prevent us from actually submitting a real move, even when a ghost piece is shown - at least that was my thinking. (On the other hand, maybe we will want to allow the move and will just handle it differently in the game logic, so nevermind all that and feel free to delete those supetfluous lines)

tsevasa: I admit this is a bit half-baked, because I tried to anticipate the next steps for the premove implementation and forgot about these lines being superfluous for now.

Naviary2: Alright! I added a new line, slightly modified, to accomplish this purpose!

Naviary2:
Ready to merge!

PR #139, opened 2024-07-29
fixed source of crash in hightlights.js

tsevasa:
Currently, `concatData_HighlightedMoves_Sliding()` takes `coords` and `legalMoves` as an input. However, the methods `concatData_HighlightedMoves_Sliding_Horz()`, `concatData_HighlightedMoves_Sliding_Vert()` and `concatData_HighlightedMoves_Diagonals()` do not use `legalMoves` at all, and instead recalculate the legal moves based on the currently selected piece, which does not make sense and causes issues when we want to show the legal moves of a piece not currently selected. I remedied that.


tsevasa:
I reversed my last commit because I was not quite sure where to move the call of `initBoundingBoxOfRenderRange()`.

PR #151, opened 2024-08-01
do not render line highlights upon arrow hover, when game is rewinded

tsevasa:


tsevasa:
Oh ok, what you did is more elegant in every way 😅

PR #154, opened 2024-08-03
Promotion fix

Idonotus:
Fixes an oversight where piece that can be promoted to but where not on the board did not have correct slide lines

### DIFF
--- a/src/client/scripts/game/chess/variant.js
+++ b/src/client/scripts/game/chess/variant.js
@@ -55,10 +55,16 @@ const variant = (function() {
      */
     function initExistingTypes(gamefile) {
         const teamtypes = new Set(Object.values(gamefile.startSnapshot.position)); // Make a set of all pieces in game
-        const rawtypes = new Set();
+        
+        // Makes sure all possible pieces are accounted for. even when they dont start with them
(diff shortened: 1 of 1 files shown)


src/client/scripts/game/chess/variant.js
```diff
@@ -55,10 +55,16 @@ const variant = (function() {
      */
     function initExistingTypes(gamefile) {
         const teamtypes = new Set(Object.values(gamefile.startSnapshot.position)); // Make a set of all pieces in game
```
Naviary3, line 57: Can this make use of gamefile.startSnapshot.existingTypes? Or is that not defined yet by this point?

Naviary3:
It's worth asking if gamefile.startSnapshot.existingTypes should also be update to include all possible promotions. What instances in the code is it used in? And does it also need to include possible promotions?

Idonotus:
This change already adds promotions to `existingTypes`. This function is responsible for init `existingTypes`. Is that what you were asking? It seems to only be used for `slidingPossible` but i think with some changes it ***could*** be used to optimize adjacency checks

Naviary3:
> This function is responsible for init `existingTypes`

Oh right, of course, shoulda read closer, I think I thought it was responsible for calculating slidingPossible

Naviary3:
Thanks!

PR #153, opened 2024-08-02
Practice Mode

tsevasa:
The goal is to introduce a new practice mode where people can practice endgames and try to beat all possible endgames in the [endgame theory table](https://docs.google.com/spreadsheets/d/14UMBn9wHz4BwGAckl4VrTFQJyzJLUw6M16dnqe-8HuU/edit?gid=0#gid=0).

Tasks:
- [x] Create guipractice.js, which is used to handle a practice menu that can create local games.
- [x] Finish Practice Menu GUI, allowing people to launch local games from a predefined list of positions, trying to checkmate a lonely black king.
- [x] Add very basic engine opponent that makes moves for the black king.
- [ ] Store player's progress in the practice mode, so that they can see what variants they have already beaten.
- [ ] Make GUI pretty.
- [ ] Make engine strong.

PR #158, opened 2024-08-04
rework insufficientmaterial.js

tsevasa:
Rework of insufficientmaterial.js. The previous version was very convoluted and inconvenient for extending, so I reworked it from the ground up. Now it fully supports all planned variants of the practice mode (#153).


Naviary2:
Thanks!

PR #161, opened 2024-08-04
Complete base code for the engine for Practice Mode

NoobJsPerson:
This PR finishes `engine.js` to make it an engine designed for use in Practice Mode #153 that can calculate to a given depth and give a final move. Its not compatible with Practice Mode engine handler and works now as a standalone script.
It features a evaluation function `loneBlackKingEval` that considers king's legal moves, pieces attacked, sum of distance from other pieces and a huge reward for capturing enemy pieces (since that usually ends in a draw by insufficient material), but the constants need to be fine tuned to improve the bot.
Also it has a tree search algorithm `negamax` you can read more about it [here](https://www.chessprogramming.org/Negamax). we use that algorithm with [alpha beta pruning](https://www.chessprogramming.org/Alpha-Beta) to speed up the search. More complicated and better search algorithms could be implemented after we improve the evaluation a bit for them to really shine.
If all moves lead into checkmate (checkmate is forced). the engine picks a random move.


Naviary2:
Cool! Can you add descriptions for each of these methods in their jsdoc for me?

NoobJsPerson:
I did. tell me if any of the descriptions are ambiguous.

PR #166, opened 2024-08-05
Add Engine to Practice Mode

NoobJsPerson:


tsevasa:
Superseeded by #167, please rename your the enginefile instead of modifying engine.js

NoobJsPerson:
I've renamed my engine file to `engineManualEval.js` is that ok?

PR #177, opened 2024-08-08
Warning when leaving active game

jermOSS:
Displays warning when trying to leave active game.

### DIFF
--- a/src/client/css/play.css
+++ b/src/client/css/play.css
@@ -778,8 +778,8 @@ button.join-button, button.copy-button {
     display: flex;
     box-shadow: 0px 1px 7px 0px rgba(0, 0, 0, 0.659);
     background: linear-gradient(to top, rgba(255, 255, 255, 0.104), rgba(255, 255, 255, 0.552), rgba(255, 255, 255, 0.216));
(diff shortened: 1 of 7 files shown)


Naviary2:
These are the changes that need to be made before I can merge:

* Merge the 'beforeunload' event listener with the one already declared in main.js
* Only add the popup in online games (not local or engine).
* Use gamefileutility.isGameOver() instead of directly reading the gameConclusion

Naviary2:
The current `beforeunload` event listener can be changed to an `unload` event, then you can add a separate `beforeunload` event listener for asking the user if they want to leave.

jermOSS:
Sadly `unload` event is deprecated.

Naviary3:
> Sadly `unload` event is deprecated.

Ah, gotcha. Here's what I'd do then. I'd experiment with removing the line that closes the socket, and see if the socket remains connected even with the popup visible. If it stays connected then great! We can adapt without the line, the socket closure message will just be different.

Naviary2:
Closing from inactivity. If you revisit this in the future, feel free to open another PR!

PR #222, opened 2024-08-19
Modularize Game logic

Idonotus:
Turn game files into ESM. This makes dependencies explicit

## Todo - required
- [x] Make every module an ESM
- [x] Allow for non-ESM communication
- [x] Automatic ESM bundling
- [x] Clean up

## Todo - clean up
- [x] **Correct imports** (very important that this is done so it wont crash)
- [x] Interfile typedefs
- [x] `HTMLScriptInjector.js`
- [x] `build.js`
- [x] `play.ejs`


Naviary2:
Removing cyclic dependancies is something that is going to take eons. Let's focus on that very last, and in separate PRs. For us to fully commit to this ESM conversion change, we first need to verify:

1. The ESMs are able to communicate with other, non-ESM scripts, like memberHeader.js.
2. That we can modularize game scripts for use both on the server and client-side, keeping the script only in one place, without breaking cross-file intellisense between other server scripts and other client scripts.
3. Our bundler is able to compact all the game scripts down to 1 classical script, for production mode.

If all of these can't be fulfilled I'd hate for you to waste a lot of work on this.

Idonotus:
> The ESMs are able to communicate with other, non-ESM scripts, like memberHeader.js.

It does work but it breaks intellisense currently

Idonotus:
> It does work but it breaks intellisense currently

Nvm intellisense is already broken so, thats annoying

Idonotus:
> 3\. Our bundler is able to compact all the game scripts down to 1 classical script, for production mode.

I have made a commit that does this

PR #249, opened 2024-09-07
Added news post regarding Classical tourney (untested)

Naviary3:


tsevasa:
Very nice! Maybe the news post should also contain text on the draw offer & new language additions? Or should that rather be a different news post?

Naviary2:
> Very nice! Maybe the news post should also contain text on the draw offer & new language additions? Or should that rather be a separate news post?

Good point. I'll just prep to push the whole update.

I see a drawback with the current language foundation, where I cannot add a news post without pushing an update. It'd be much nicer if posts/blogs were individual documents, written in markup, that could easily be drag-and-dropped into a blog folder of some sort 🤔

PR #251, opened 2024-09-17
Reformat modules

Idonotus:
Standardize module format through flattening
Should not change functionality or features

### DIFF
--- a/src/client/scripts/game/chess/movesets.js
+++ b/src/client/scripts/game/chess/movesets.js
@@ -6,203 +6,198 @@
  * 
  * ZERO dependancies
  */
-const movesets = (function() {
 
-    /**
-     * Returns the movesets of all the pieces, modified according to the specified slideLimit gamerule.
-     * 
-     * These movesets are called as functions so that they return brand
-     * new copies of each moveset so there's no risk of accidentally modifying the originals.
-     * @param {number} slideLimit - Optional. The slideLimit gamerule value.
-     * @returns {Object} Object containing the movesets of all pieces except pawns.
-     */
-    function getPieceMovesets(slideLimit = Infinity) {
-        if (typeof slideLimit !== 'number') throw new Error("slideLimit gamerule is in an unsupported value.");
+/**
+ * Returns the movesets of all the pieces, modified according to the specified slideLimit gamerule.
+ * 
+ * These movesets are called as functions so that they return brand
+ * new copies of each moveset so there's no risk of accidentally modifying the originals.
+ * @param {number} slideLimit - Optional. The slideLimit gamerule value.
+ * @returns {Object} Object containing the movesets of all pieces except pawns.
+ */
+function getPieceMovesets(slideLimit = Infinity) {
+    if (typeof slideLimit !== 'number') throw new Error("slideLimit gamerule is in an unsupported value.");
 
-        return {
-            // Finitely moving
-            pawns: function() {
-                return { individual: [] };
-            },
-            knights: function() {
-                return {
-                    individual: [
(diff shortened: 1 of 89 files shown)


src/client/scripts/game/chess/movesets.js
```diff
@@ -6,203 +6,198 @@
...
+                    // + Knight moveset!
+                    [-2,1],[-1,2],[1,2],[2,1],
+                    [-2,-1],[-1,-2],[1,-2],[2,-1]
+                ]
+            };
+        },
+    };
+}
 
-export default movesets;
```
Naviary3, line 208: Is the Piece type definition exported?

src/client/scripts/game/chess/organizedlines.js
```diff
@@ -24,342 +24,358 @@ import coordutil from '../misc/coordutil.js';
...
+    organizePiece,
+    removeOrganizedPiece,
+    initUndefineds,
+    appendUndefineds,
+    areWeShortOnUndefineds,
+    addMoreUndefineds,
+    insertUndefinedIntoList,
+    buildKeyListFromState,
+    buildStateFromKeyList,
+    getEmptyTypeState,
```
Naviary3, line 376: getEmptyTypeState() will only ever be called from inside this module, so it doesn't need to be exported.

src/client/scripts/game/chess/specialmove.js
```diff
@@ -9,116 +9,121 @@ import coordutil from '../misc/coordutil.js';
...
-
-export default specialmove;
\ No newline at end of file
+export default {
+    getFunctions,
+    kings,
+    pawns,
+    isPawnMoveADoublePush,
+    getEnPassantSquare,
+    getEnpassantCaptureCoords,
```
Naviary3, line 128: getFunctions() is the only method called externally, all others don't need to be exported, but kept private.

src/client/scripts/game/chess/specialundo.js
```diff
@@ -10,113 +10,116 @@ import coordutil from '../misc/coordutil.js';
...
-};
+    return true; // Special move has been undo'd!
+}
 
-export default specialundo;
\ No newline at end of file
+export default {
+    getFunctions,
+    kings,
+    pawns,
```
Naviary3, line 123: getFunctions() is the only method called externally

src/client/scripts/game/gui/stats.js
```diff
@@ -16,111 +16,125 @@ import config from '../config.js';
...
+    showPiecesMesh,
+    updatePiecesMesh,
+    hidePiecesMesh,
+    showFPS,
+    hideFPS,
+    updateFPS,
+    showRotateMesh,
+    updateRotateMesh,
+    hideRotateMesh,
+    hideMoveLooking
```
Naviary3, line 139: hideMoves() is not called externally, neither:
setTextContentOfMoves,

and hideMoveLooking() is depricated, may as well delete.

Naviary3:
@Idonotus I just considered the possibility that flattened modules might be broken in production mode, as the builder has to compile them all into one script, which may lead to a bunch of duplicate functions. But, maybe the builder is smart enough to avoid that by isolating the module. Have you tested production mode?

Idonotus:
@Naviary3 I did some testing
1) It renames functions to avoid collisions
2) It "simulates" exports to avoid changing everything else

It has been working and i don't foresee this being a problem

Naviary2:
Thanks! Looks good to me! One less level of indentation everywhere will be nice.

Naviary2 (approved):

PR #254, opened 2024-09-22
Update chinese translation

Heinrich-XIAO:


Heinrich-XIAO:
The chinsee translation is UTD.

Naviary2:
Thank you!

PR #256, opened 2024-09-23
Traditional Chinese Translation

Heinrich-XIAO:


Heinrich-XIAO:
Translaton done. The only real difference between traditional and simplified chinese is a difference in some of the characters.

Naviary2:
There seems to be only 146 lines in the Traditional Chinese file, where I was expecting 600+ to match the others

translation/zh-TW.toml
```diff
@@ -357,7 +357,7 @@ please_report_bug = "這不應該發生，請報告此錯誤！"
 checkmate = "將死"
 stalemate = "僵局"
 repetition = "三次重復"
-moverule = [, "-回合規則"]  # The game inserts a number inbetween these two strings
+moverule = ["", "-回合規則"]  # The game inserts a number inbetween these two strings
```
Naviary2, line 360: I meant to say missing quotes, not parenthesis

Naviary2:
Thank you!

PR #261, opened 2024-09-29
Replace string piecetypes with int piecetypes

Idonotus:
Replace string piecetypes with int piecetypes. Adds some pre-multiplayer piece colouring compatibility as well.

## Todo
- [ ] Fix pawn moves and promotions
- [ ] Centralize piece order
- [ ] Consider merging `colorutil` and `typeutil`
- [ ] Clean up


Naviary2:
This frightens me a little. In my head we were going to create a new universal gamefile format first, one that can be used on both client and server-side, and that would be the format that uses integers to stire piece types.

I haven't studied the changes you've made so far, but potentially this stuff would get overwritten or be no longer needed once we have a univefsal gamefile format.

Idonotus:
Yea that's fair. I thought we had agreed on keeping the organized lines and using int types for it though?

Naviary2:
> Yea that's fair. I thought we had agreed on keeping the organized lines and using int types for it though?

Yes. [Here's a rough outline of ideas we can try to optimize the new gamefile format](https://discord.com/channels/1114425729569017918/1115358966642393190/1282153348405137480).

Idonotus:
It doesnt seem that in pieces save that much memory from my look at the memory tool. I'll close this anyway until the format has been finalised

PR #262, opened 2024-09-29
Update vulnerable dependencies

this-is-not-available:
From 6 vulnerabilities (3 high, 3 moderate severity) to 0
Also make package name correct (no uppercase and no spaces)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "Infinite Chess Server",
+  "name": "infinite-chess-server",
   "version": "1.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


Heinrich-XIAO (commented):
The name was changed. Is that intentional?

this-is-not-available:
Probably. It makes sense though

Heinrich-XIAO:
Also, is it still 1.0.0?

Heinrich-XIAO:
Yeah.

this-is-not-available:
Done

Naviary2:
Thanks! I'll merge this.

It would be nicer if there was just one location we could keep the server version, instead of 4.. these 2 I probably won't always update

Issue #258, opened 2024-09-28
Using tabs instead of spaces


Heinrich-XIAO:
I think we should use tabs instead of spaces as tabs can look different based on preferences. In this case, consistency is not ideal as different programmers want different tab sizes. I use 2 spaces per tab while the code in the repo all use 4 spaces per "tab".

Heinrich-XIAO:
Github also renders tabs like this, so consistency between tabs and spaces is important. 
![image](https://github.com/user-attachments/assets/604283a7-1354-462a-b87e-c0bc8611b39f)

Heinrich-XIAO:
Github allows you to change tab size, but this only works when they are real \t characters 
![image](https://github.com/user-attachments/assets/b9243b5f-323d-47bb-88e3-dd992644ab92)

Naviary2:
Valid arguments.

When you say tabs can look different depending on your preferences, are you only referring to the length of tabs, or are there additional customizations for tabs?

I do think the way github renders tabs is a good reason to switch, but you aren't actually having to hit your space bar 4 times every line are you? My tab auto-inserts 4 spaces.

Heinrich-XIAO:
probably only the length on most editors, but in vim-like editors, I'm pretty sure I can change colour as well.

Heinrich-XIAO:
> but you aren't actually having to hit your space bar 4 times every line are you? My tab auto-inserts 4 spaces.

Here in neovim, I do have to slam space unless I configure tab to do that for me, but I can't because all of my other projects use \t.

Naviary2:
Adopted! All indentation in the project is now tabs instead of spaces.

PR #264, opened 2024-09-30
Fix indentation everywhere else

Heinrich-XIAO:
Indentation didn't seem to be fixed in a lot of places, so this pr should fix that.


Naviary2:
Interesting, so it missed some TOMLs? There should be an automatic way to fix all indentation issues, I don't want you to spend all the time to manually fix each line. I suppose you could also search for every group of 4 spaces and mass-replace that way.

Heinrich-XIAO:
> I suppose you could also search for every group of 4 spaces and mass-replace that way.

Some of the files won't work. like license being the simplest to fix. i think it needs some regex or something

Naviary2:
Some files we could probably ignore, like ones that will never be modified. For example, I skipped modifying gl-matrix, which was a resource downloaded from online

Heinrich-XIAO:
Why is #258 closed if this pr isn't?

Naviary2:
There was just a few straggler files that still contained spaces for indentation, right? Is this one ready for review?

Heinrich-XIAO:
I think so. I've started using the draft and open pr statuses correctly.

translation/changes.json
```diff
@@ -1,145 +1,146 @@
...
+	"17": {
+		"note": "Added credits.language_heading and credits.language_credits, line 82",
+		"changes": []
+	},
+	"16": {
+		"note": "Deleted all news posts. The only keys you keep in the `news` section is `title` and `more_dev_logs`! Please verify your language's news posts have been translated correctly within translation/news!",
+		"changes": []
+	},
+	"15": {
+		"note": "Added news.sept9-2024, line 493",
```
Naviary2, line 11: Could you revert this?

Naviary2:
Thanks!

PR #272, opened 2024-10-04
Update french translation

Heinrich-XIAO:
I speak french, but it's canadian french, but in this case  it's probably fine.


Naviary2:
I will note, even though the french TOML is now up to date, there is still 1 untranslated news post

PR #282, opened 2024-10-21
Refactor clock.js

Idonotus:
Separated the gui clock functionality into a different script.
Moved clock game data into gamefile.

### DIFF
--- a/src/client/scripts/game/chess/gamefile.js
+++ b/src/client/scripts/game/chess/gamefile.js
@@ -172,6 +172,32 @@ function gamefile(metadata, { moves = [], variantOptions, gameConclusion } = {})
 	/** Contains the methods for undo'ing special moves for this game. */
 	this.specialUndos = undefined;
 
+	/** The time each player has remaining, in milliseconds. */
+	this.currentTime = {
+		white: undefined,
+		black: undefined,
+	};
+
+	/** Contains information about the start time of the game. */
+	this.startTime = {
+		/** The number of minutes both sides started with. */
+		minutes: undefined,
+		/** The number of miliseconds both sides started with. */
+		millis: undefined,
+		/** The increment used, in milliseconds. */
+		increment: undefined,
+	};
+
+	/** Which color's clock is currently running. This is usually the same as the gamefile's whosTurn property. */
+	this.colorTicking = undefined;
+	/** The amount of time in millis the current player had at the beginning of their turn, in milliseconds. */
+	this.timeRemainAtTurnStart = undefined;
+	/** The time at the beginning of the current player's turn, in milliseconds elapsed since the Unix epoch. */
+	this.timeAtTurnStart = undefined;
+	this.timeNextPlayerLosesAt = undefined;
+	/** True if the game is not timed. @type {Boolean}*/
+	this.untimed = undefined;
+
 	// JSDoc stuff over...
 
 	// this.metadata = metadata; // Breaks the above JSDoc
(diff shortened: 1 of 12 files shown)


Naviary2:
Screen Shot 2024-10-21 at 4 11 38 PM
Functionality-wise, this seems to be the only issue. The clocks are lit up on the first 2 moves when they should only be when they are active.

src/client/scripts/game/chess/gamefile.js
```diff
@@ -172,6 +172,32 @@ function gamefile(metadata, { moves = [], variantOptions, gameConclusion } = {})
 	/** Contains the methods for undo'ing special moves for this game. */
 	this.specialUndos = undefined;
 
+	/** The time each player has remaining, in milliseconds. */
+	this.currentTime = {
+		white: undefined,
+		black: undefined,
```
Naviary2, line 178: All of these clock-related variables can you place together inside an object say, "clock"? Similar to how mesh data is all together.

src/client/scripts/game/misc/timeutil.js
```diff
@@ -152,6 +152,49 @@ function isCurrentDateWithinRange(startMonth, startDay, endMonth, endDay) {
 		&& checkDate <= new Date(checkDate.getFullYear(), endMonth - 1, endDay);
 }
 
+function getTextContentFromTimeRemain(time) {
+	let seconds = Math.ceil(time / 1000);
+	let minutes = 0;
```
Naviary2, line 157: I think the new methods added to timeutil.js could be moved to clock.js. timeutil.js I intended mainly for working with dates

Idonotus: https://github.com/Infinite-Chess/infinitechess.org/pull/282/commits/8aa01c2c05a1b8af2597241946e6cbf1aaa3159
Clock data is now contained under `gamefile.clocks`

src/client/scripts/game/chess/gamefile.js
Idonotus: https://github.com/Infinite-Chess/infinitechess.org/pull/282/commits/8aa01c2c05a1b8af2597241946e6cbf1aaa3159
Clock data is now contained under `gamefile.clocks`

src/client/scripts/game/misc/timeutil.js
Idonotus: I'd say keep it like this to reduce dependency hell, maybe move `getMinutesAndIncrementFromClock` to clock.js as it is only used by clock.js

Heinrich-XIAO (approved):

src/client/scripts/game/misc/timeutil.js
Naviary3: Perhaps we should then be reducing the dependancies of clock.js? `game` is only used for `game.getGamefile()` so we can solve that by passing the gamefile as a parameter. `onlinegame` is also not needed if we add a parameter.

But, if we want to hold off on resolving (some) of those dependancies for now, I still feel these new methods in `timeutil.js` don't fit right. Like, they only make sense with the context that they are working with gamefile clocks, whereas all the other methods in the script work with dates, UTC time, conversion from milliseconds to minutes and other generalized stuff. If you'd rather, do you want to make a new script `clockutil.js` with zero dependancies that stores the conversion methods related to gamefile clocks?

Idonotus: I see what you mean. I've moved it into `clockutil.js` for now

Naviary2: Thanks!

Naviary2:
Screenshot 2024-10-23 at 2 52 06 PM

Placing this here for documentation

Heinrich-XIAO (approved):

Heinrich-XIAO (approved):

Naviary2:
Thanks! Working great now!

PR #283, opened 2024-10-22
Refactoring of the header implementation (DRAFT)

### LINKED ISSUE #12: Change header and footer implementations
Currently the header and the footer are hardcoded in all the html pages that need them. It is not that big of an issue since they are not often changed but it would be better to have them written only once, especially considering that the number of pages will probably grow over time

### LINKED ISSUE #278: Move language selector to page header
The way that it is now, it is likely that people will miss it, because it is tucked down at the bottom of the page.

xa31er:
Moves the header to header.ejs and creates header.css
- Closes #278 
- Closes #12

### DIFF
--- a/translation/zh-CN.toml
+++ b/translation/zh-CN.toml
@@ -1,6 +1,6 @@
 name = "简体中文" # Simplified Chinese - Name of language
 direction = "ltr" # Change to "rtl" for right to left languages
-version = "17"
+version = "18"
 maintainer = "Heinrich Xiao"
 
 [header]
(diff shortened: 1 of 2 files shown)


Naviary3:
Oh cool you're starting on this! Please post common screenshots so we can give feedback on how it looks!

My idea for the overhauled header is:
* Site logo & name on the very left.
* No more horizontal scrolling needed on small screens, turn the words into vector graphics so each of the links fit on the bar.
* Add a globe svg for selecting language?
* Also keep in mind that I am soon planning on adding a gear-icon on the very right for a settings dropdown menu.

Naviary3:
This could even be split into two PRs if you want to merge the part now that turns the header into a component, as it looks like that is already done. We just need a little javascript for each page that will grey-out the header link if it is the current page we're on.

xa31er:
I'm going to split it

PR #285, opened 2024-10-23
Refactoring of header implementation.

### LINKED ISSUE #12: Change header and footer implementations

xa31er:
- Closes #12 (that's an old one)
- Bonus! Improved language selector legibility.

### DIFF
--- a/src/client/css/createaccount.css
+++ b/src/client/css/createaccount.css
@@ -193,13 +193,6 @@ a {
 
 /* Start increasing header links width */
 @media only screen and (min-width: 450px) {
-    header {
-        overflow: unset;
-    }
-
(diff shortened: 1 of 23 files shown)


xa31er:
I missed the current page highlight. Gotta add it.

xa31er:
Now that I look at it, it was a very rushed PR, 
I'll try to be more thorough with the modification.

Naviary2:
Nah it was a good start! Just some stuff missing 😆

PR #292, opened 2024-10-29
Add Elo ratings

just-another-nerd:


just-another-nerd:
Sorry about the huge diffs. package-lock.json changed a lot it seemed

this-is-not-available:
This looks good! However, we settled at a dynamic K factor and at separate elo systems for the distinct variants there are https://discord.com/channels/1114425729569017918/1260310049889189908/1261140553068707872 . Also we shouldn't add elo until we make the game more infinite https://discord.com/channels/1114425729569017918/1260310049889189908/1261147187941998602

Naviary2:
Good start! But yeah, unfortunately we're waiting on a couple other features before we implement this. They are now listed here #295

PR #298, opened 2024-11-02
Rate limiter no longer uses timers

Naviary3:


Naviary3:
Ah I only meant to pull the changes from the rate limit script

Naviary2:
The rate limiting change has been applied to the main branch

PR #299, opened 2024-11-06
Cleaned up css, bufferdata.js, shapes.js, and legalmoveshapes.js

Naviary3:
NOT TESTED for crashes. Must fix all errors and make sure all rendered items haven't been broken.


Naviary3:
Also make sure all rendered items still work just beyond 10,000, to work with "offset".

PR #315, opened 2024-11-11
Reorganization of logic files

Idonotus:
Move logic files from `esm/game` into `esm/chess` and move the util files to designated folders

# Todo
- [x] Separate logic files from game
- [ ] Come up with non duplicate names
- [ ] Correct type imports

### DIFF
--- a/src/client/scripts/esm/components/header/themes.js
+++ b/src/client/scripts/esm/components/header/themes.js
@@ -16,7 +16,7 @@ const defaults = {
 	neutralPiecesColor: [1, 1, 1, 1],
 };
 
-const defaultTheme = 'sandstone';
+const defaultTheme = 'wood_light';
 
 const themeDictionary = {
 
--- a/src/client/scripts/esm/game/chess/backcompatible.js
+++ b/src/client/scripts/esm/game/chess/backcompatible.js
@@ -1,0 +1,171 @@
+
+// Import Start
+import formatconverter from './formatconverter.js';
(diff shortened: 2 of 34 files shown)


Naviary3:
Looks really good! For the todo, what duplicate names do you mean?

My only change would be:
* movesscript.js contains some gui logic, so it may not be the best fit in chess logic (outside of the game directory).

A lot of imports have changed, have you extensively tested for new bugs that may have been introduced? Below is many things I would test with a change this big. For each of these scenarios, please make sure all the GUI elements behave correctly, and that there's no unexpected errors in the client-console nor the server console.

* Online games, does checkmate end the game as expected? What about offering draws?
* Local games
* Timed games
* Untimed games
* Pasting games
* Visiting every page
* logging in
* logging out
* account creation
* account deletion
* anything more you think of...

Once this is merged, the chess logic directory will step up on priority to have a healthy dependancy tree, because that directory we want the server to be able to use too. The game directory is less important, but could come last.

Idonotus:
> Looks really good! For the todo, what duplicate names do you mean?

there are two folders called util and chess it may cause some issue when talking about both contexts

Idonotus:
Thankfully by using debugging with the bundler all import issues have already been squashed within game.

I have tested so far:
- AFK kick
- Draw offer
- Untimed
- Timed 1m 2m
- Sounds
- Pasting and move viewer
- Theme changes
- Language changes
- Log in/out creation/deletion
- Stalemate
- Insufmat
- Rejoining

PR #316, opened 2024-11-12
Update polish translation to version 22

Apsurt:
I've also made the changes field in changes.json more verbose and added line indicators to the note.


Naviary3:
Hi! Thanks @Apsurt! I will be more specific about what lines changed in the note.

However, the "changes" field actually has an effect on how the website loads other languages: Whatever key is inside will get auto-replaced by the english key if the language is not on the latest version. A byproduct is, by specifying a group like "header", *every single word* in the header will then be in english for every language that's behind, EVEN though most of the words didn't change. There's not many cases where this is useful, which is why I usually leave it blank, because the behind-translation is almost always preferred over defaulting to english.

Naviary2:
I am no longer pushing changes to this branch, but the new themes-database instead. You alright if I copy over your changes, or do you want the commit credit?

If you do you want the commit credit have to create a new branch on your fork, sourced from themes-database, make the changes there and re-PR.

Naviary2:
I've copied the specific changes over

PR #327, opened 2024-11-17
Added 5 dimensional variant

frankm1870:


Idonotus (changes requested):
Maybe add the .vs directory to the .gitignore so it wont be an issue

src/client/scripts/esm/chess/logic/fivedimensional.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
```
Naviary2, line 1: Can you rename the file to include promote in its name since that's the only thing it does?

src/client/scripts/esm/chess/variants/fivedimensionalgenerator.js
```diff
@@ -0,0 +1,103 @@
...
+
+// Currently board spacings other than 10 are not supported by the position generator, but are supported
+// by the moveset generator.
+const BOARD_SPACING = 10;
+
+function genPositionOfFiveDimensional() {
+	const standardPosStr = 'P1,2+|P2,2+|P3,2+|P4,2+|P5,2+|P6,2+|P7,2+|P8,2+|p1,7+|p2,7+|p3,7+|p4,7+|p5,7+|p6,7+|p7,7+|p8,7+|R1,1+|R8,1+|r1,8+|r8,8+|N2,1|N7,1|n2,8|n7,8|B3,1|B6,1|b3,8|b6,8|Q4,1|q4,8|K5,1+|k5,8+';
+
+	const standardPos = formatconverter.ShortToLong_Format(standardPosStr).startingPosition;
+	const result = { ...standardPos };
```
Naviary2, line 18: Why do we need to copy the position into a new object?

src/client/scripts/esm/chess/variants/variant.js
```diff
@@ -359,10 +370,54 @@ function getApplicableTimestampEntry(object, { UTCDate, UTCTime }) {
 	return object[timestampToUse];
 }
 
+/**
+ *
```
Naviary2, line 375: Please add a description

Issue #23, opened 2024-07-10
Board momentum


optim-ally:
I noticed that board navigation has momentum while zooming in/out but not when moving side-to-side/up-and-down. Should the board also have translational momentum?

(By "momentum" I mean that it continues to move after the user has stopped interacting and gently slows down until it comes to a halt.)

Naviary2:
It does! But only when using `WASD` to move, not with the mouse.

Entificationcurry:
They sent me to work on it 🔥(that's how you know your suggestion will never occur, it took me 10 hours to make a glitchy version 😭)

Entificationcurry:
Well nevermind I finished it and it works pretty well

Naviary2:
@optim-ally https://discord.com/channels/1114425729569017918/1115358966642393190/1305278858978656277

Naviary2:
This is now a feature on desktop! I won't close yet as it isn't applied to mobile devices that don't support a mouse, but an idea I have for getting it to work is:

Whenever you drag the board, whether it's by the mouse or by your finger, the boards position changes, right? Well if we just record the recent history of the board position, we can pull its velocity directly out of that! We constantly update its velocity while it's being dragged, and even though it's snapped to the mouse/finger, once you release it that velocity will take over!

BF5258:
> if we just record the recent history of the board position, we can pull its velocity directly out of that!

Do you mean Verlet integration?

Naviary2:
So BF5258 is holding out on us and sneakily added this board momentum feature on mobile and didn't tell us.
https://github.com/BF5258/infinitechess.org/tree/dragging-pieces

BF5258 if you can see this, please separate it from the unrelated code and PR 🙂

Naviary2:
Thank you BF5258! #337 Has marked this as complete, working on both desktop and mobile!

PR #337, opened 2024-11-20
Added board momentum for touchscreen.

BF5258:
#23


BF5258:
Does the position history need to be cleared by setBoardPos?  Instantaneous teleportation doesn't seem to be used in game currently, but if it were added in the future, the board could get thrown at ridiculous velocities after teleporting.

Naviary2:
Thanks so much BF5258!!! 🎉 That's actually really nice how it throws the scale as well, it feels really good!

I made a couple minor changes to make it PR ready, and increased friction a little bit. As for when to clear the position history- I just cleared it whenever the board is dropped, that should cover our needs for a while. I'm satisfied with where this is at now!

Closes #23 
(this phrase should hopefully automatically close that issue once I merge this)

PR #338, opened 2024-11-20
Added file to move members.json to the database

frankm1870:


Naviary2:
I don't think it's fully working yet, but I'm going to merge it so I can get it working afterward, since there's changes on the branch.

Issue #22, opened 2024-07-10
Dragging pieces


optim-ally:
Dragging the board is a new kind of user interaction in infinite chess vs 8x8 chess. It makes the familiar action of dragging pieces in online chess problematic. I tried several time to drag pieces in infinite chess before remembering that it didn't work like that, but perhaps it could!

We could introduce ways to explicitly choose "piece dragging" or "board dragging", like right clicking or holding down a particular key, but I expect most users won't bother with those.

I propose the following logic for handling "drag" interactions:

1. if the board is already moving (e.g. it has remaining momentum from being dragged), then drag the board
2. else if the click action was a right-click or CMD/CTRL was held, drag the board
3. else if the cursor is not over a piece of the player's colour, drag the board
4. else if the cursor is over multiple pieces (when zoomed far enough out), drag the board
5. else drag the piece

This might need to go through several iterations if it's going to work, but what are initial thoughts?

Heinrich-XIAO:
I feel like a good solution to this would be to drag the piece if the mouse was directly on the piece and drag the board otherwise.

optim-ally:
Essentially, yes, but we need to be careful with the details. For example, if the board is zoomed out far enough then the mouse may be on top of multiple pieces.

Another issue is whether we consider the mouse "on" a piece if it is in the same square or if it actually needs to be inside the rendered shape of the piece. If it's the latter then cosmetic changes to the board and pieces could affect playing experience by making pieces harder or easier to move. I don't think that should be the case.

Naviary2:
> 1. if the board is already moving (e.g. it has remaining momentum from being dragged), then drag the board

When dragging the board with the mouse and letting go, it no longer has momentum, it stops immediately. So it can't be *thrown* in this manner.

> if the board is zoomed out far enough then the mouse may be on top of multiple pieces.

The coordinates of the mouse is a single point, so it can never technically be over multiple pieces, at least in the code. But at that zoom distance clicking and dragging probably shouldn't be selecting a piece anyway, the user probably intends on moving the board.

> I feel like a good solution to this would be to drag the piece if the mouse was directly on the piece and drag the board otherwise.

I like this solution and I have actually thought this before! When I check out Lichess.org's mouse controls, they also have click to select, and dragging as well! Should have thought of that when I was making the first release! I personally think this is the best solution. But it can be tested!

optim-ally:
[...]

PR #350, opened 2024-11-29
Dragging pieces

BF5258:
Implements dragging pieces but not interaction with arrow indicators or auto-pan.
Issue #22


Naviary2:
Hey nice! Do you want me to review this as it is now (I know it's unfinished) and provide feedback?

BF5258:
Yes, that's fine. Though, I will be without WiFi until sometime next week.

src/client/scripts/esm/chess/logic/movepiece.js
```diff
@@ -55,11 +55,12 @@ import game from '../../game/chess/game.js';
  * - `doGameOverChecks`: Whether to perform game-over checks, such as checkmate or other win conditions.
  * - `concludeGameIfOver`: If true, and `doGameOverChecks` is true, then if this move ends the game, we will not stop the clocks, darken the board, display who won, or play a sound effect.
  * - `animate`: Whether to animate this move.
+ * - `animateSecondary`: Animate the pieces affected by the move without the piece that made the move. Used after dragging the king to castle.
```
Naviary3, line 58: I like the description you added! It's just unclear if it overwrites `animate` or if that overwrites `animateSecondary`

src/client/scripts/esm/game/chess/selection.js
```diff
@@ -39,8 +41,27 @@ import config from '../config.js';
  * including the legal moves it has available.
  */
 
+/**
+ * Bugs:
+ * - Fixed I think?
+ * To do:
+ * - Allow dragging the board with a second finger while dragging a piece.
```
Naviary3, line 48: That's interesting behavior! Could be good. Have you also thought of potentially terminating a drag if a 2nd finger is present? Like, I can imagine you go to zoom out with your fingers and one finger touches down a split second before the other, causing you to drag a piece with that finger and pan with the 2nd finger, instead of zooming out. What are your thoughts?

src/client/scripts/esm/game/chess/selection.js
```diff
@@ -39,8 +41,27 @@ import config from '../config.js';
...
+/**
+ * When dropped in the same square, pieces are unselected every second time.
+ * This alows players to move pieces by clicking if they don't want to use drag.
+ * @type{boolean} 
+ * */
+let didLastClickSelectPiece;
+/** Set to false if the user wants to use the original click controls. @type {boolean} */
+let dragEnabled = true;
+/** Is the user using touchscreen. If so we should ignor mouse over. @type {boolean} */
+let touchscreenMode = false;
```
Naviary3, line 64: docutil.js has a method for determining if the screen is touch-compatible.

Do you know about the chrome dev tools mode that lets you simulate a touch screen?

src/client/scripts/esm/game/chess/selection.js
```diff
@@ -116,27 +139,47 @@ function update() {
 		if (promoteTo) makePromotionMove();
 		return;
 	}
+	if (perspective.isLookingUp() && draggingPiece) return draganimation.hideHeldPiece(); //Don't render the draggedPiece if we are looking at the sky.
```
Naviary3, line 142: I imagine just a return statement here would do. And in the render method, replace the `hidden` variable with a check for if we're looking up, and if we are, don't regen the model or render.

src/client/scripts/esm/game/input.js
```diff
@@ -65,6 +67,10 @@ let ignoreMouseDown = false;
 
 let mouseIsSupported = true;
 
+/** Touchscreen and mouse */
+let usingTouchscreen; // True if the mouse recent pointer input was from a touch event.
```
Naviary3, line 71: This variable name is a little confusing for what it means. Maybe something like `pointerIsTouch`? Is your intention that this will work on devices that have both a mouse AND touch-screen compatibility?

src/client/scripts/esm/game/input.js
```diff
@@ -113,6 +119,30 @@ function getMouseWorldLocation() {
...
+
+function getPointerClickedTile() {
+	return usingTouchscreen ? touchClickedTile : mouseClickedTile;
+}
+
+function getPointerWorldLocation() {
+	return [pointerWorldLocation[0], pointerWorldLocation[1]];
+}
+
+function getUsingTouchscreen() {
```
Naviary3, line 142: Again, docutil provides a method for determining if the device is touch compatible, or is this function meant to just tell you whether the last pointer event was touch-activated?

src/client/scripts/esm/game/rendering/draganimation.js
```diff
@@ -0,0 +1,131 @@
...
+/** The hight the piece is rendered above the board when in perspective mode. */
+const perspectiveHeight = 0.6;
+const shadowColor = [0.1, 0.1, 0.1, 0.5];
+
+let startCoords;
+let endCoords;
+let pieceType;
+
+let transparentModel;
+let pieceModel;
```
Naviary3, line 49: Since the models are generated and rendered in the rendering methods, they don't need to be in the public scope of the script.

src/client/scripts/esm/game/rendering/draganimation.js
```diff
@@ -0,0 +1,131 @@
...
+	pieceModel = buffermodel.createModel_ColorTextured(new Float32Array(data), 3, "TRIANGLES", spritesheet.getSpritesheet());
+}
+
+/**
+ * 
+ * @param {string} type - The type of piece being dragged
+ * @param {number} pieceCoords - the square the piece was on
+ * @param {number[]} coords - the world coordinates the piece has been dragged to
+ */
+function dragPiece(type, pieceCoords, coords) {
```
Naviary3, line 100: It's strange how you use dragPiece() to both initiate and update the position of the dragged piece. Could the logic for updating be moved into a new function?

Naviary3:
Thanks for adding all of the jsdoc comments! It really helps

Naviary3:
Finished a code review, I can test the functionality later

Naviary2:
I just tested! Wow it works so well I'm getting excited! 🙌

Everything feels robust except for 2 things:

* It seems to render the dragged piece at a constant size, ignoring the current scale. Perhaps the "minimum size" is accidentally being applied 100% of the time?
* Specifically in perspepctive mode, it won't let me drag a piece to move it if I let go less than 0.5 seconds after I first picked it up. I have to move it slower for it to register. Perhaps it's something related to simulated mouse clicks? since if you let go of the mouse within a small time window it simulates a click where the mouse was originally pressed down, which is on the piece's start square?

src/client/scripts/esm/game/chess/selection.js
BF5258: touchscreenMode was a duplicate of pointerIsTouch. I forgot to remove it in commit [aeb699e](https://github.com/Infinite-Chess/infinitechess.org/pull/350/commits/aeb699e201385018e6a8051cb5c9395c46823c58).

> Do you know about the chrome dev tools mode that lets you simulate a touch screen?

No, I didn't. Thank you for the advice! It will be very helpful.

src/client/scripts/esm/game/input.js
BF5258: The variable is for devices that support both. I wanted the piece to be offset upward only if the user is currently using the touchscreen. Is there a better way to do this?

BF5258:
> On non-touch screen devices, It seems to render the dragged piece at a constant size, ignoring the current scale.

Yes, the dragged piece only scales with the board in perspective mode. Should it always scale with the board unless touchscreen is being used? If so, how should we prevent it from becoming too small to distinguish?

Naviary3:
Atleast when the piece was **smaller** than the square it looked really off..

Also, it will be easier to tell where the piece is once the rank/column is outlined while dragging. But we can always play around with some settings to see which one feels best.

BF5258:
> On mobile, I can no longer single-tap to move the pieces, nor can I select opponent pieces to view their moves.

Oops. I forgot to convert from {id, x, y} to [x, y]. It should be fixed now.
image

src/client/scripts/esm/game/input.js
Naviary2: That's reasonable then to have a variable then to store whether the current pointer is a finger or not! I'll repeat the variable name was just a little confusing.

Naviary2:
Thanks! Let me know when you want another review 👊

src/client/scripts/esm/game/chess/selection.js
BF5258: Good idea. Dragging pieces by accident could get annoying.
Is to okay if auto-pan is implemented in a future pull request?

Naviary2:
I will probably close this soon in favor of #377 which has all the same changes, and more

Naviary2:
Closing for the same reason as the above comment.

PR #351, opened 2024-11-29
Converted variant.js to typescript, and fixed typescript compilation

frankm1870:


tsconfig.json
```diff
@@ -12,8 +12,7 @@
 		"noPropertyAccessFromIndexSignature": true,
 		"noUncheckedIndexedAccess": true,
 		"strict": true,
-		"noEmitOnError": true,
-		"tsBuildInfoFile": "./dist/tsconfig.tsbuildinfo"
```
Naviary2, line 16: Umm why did these get deleted

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -1,13 +1,24 @@
 
+// @ts-ignore
```
Naviary2, line 2: All these repeated ts-ignore lines are kind of ugly and probably not sustainable for every single file. What's the issue we get without them?

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -404,12 +510,16 @@ function getMovesetsOfVariant({ Variant, UTCDate = timeutil.getCurrentUTCDate(),
  * Any piece type present in the modifications will replace the default move that for that piece.
  * The slidelimit gamerule will only be applied to default movesets, not modified ones.
  * @param {Object} movesetModifications - The modifications to the default movesets.
- * @returns {number} [defaultSlideLimitForOldVariants] Optional. The slidelimit to use for default movesets, if applicable.
+ * @param {number} [defaultSlideLimitForOldVariants] Optional. The slidelimit to use for default movesets, if applicable.
  * @returns {Object} The pieceMovesets property of the gamefile.
  */
-function getMovesets(movesetModifications = {}, defaultSlideLimitForOldVariants) {
+function getMovesets(movesetModifications: Moveset = {}, defaultSlideLimitForOldVariants = Infinity) {
```
Naviary2, line 520: defaultSlideLimitForOldVariants should be an optional argument, instead of defaulting to infinity, as movesets.getPieceDefaultMovesets() already has a default value for it when not specified.

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -33,6 +44,46 @@ const coaIPPromotionsAllowed = repeatPromotionsAllowedForEachColor(coaIPPromotio
 
 const gameruleModificationsOfOmegaShowcasings = { promotionRanks: null, moveRule: null, turnOrder: ['black', 'white'] }; // No promotions, no 50-move rule, and reversed turn order.
 
+interface Moveset {
```
Naviary2, line 47: Can't help but feel movesets.js would be a better place to store this interface

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -33,6 +44,46 @@ const coaIPPromotionsAllowed = repeatPromotionsAllowedForEachColor(coaIPPromotio
...
+	[key: string]: {
+		individual: number[][],
+		sliding?: {
+			[key: string]: number[]
+		}
+	}
+}
+
+interface Gamerules {
```
Naviary2, line 56: Same thought as above, but gamerules.js intead of movesets.js

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -242,7 +299,17 @@ function getStartingPositionOfVariant({ Variant, UTCDate, UTCTime }) {
  * & pawnDoublePush & castleWith properties.
  * @returns {Object}
  */
-function getStartSnapshotPosition({ positionString, startingPosition, specialRights, pawnDoublePush, castleWith }) {
+function getStartSnapshotPosition({ positionString, startingPosition, specialRights, pawnDoublePush = false, castleWith }: {
```
Naviary2, line 302: If `pawnDoublePush` is an optional argument, does it need a default value?

tsconfig.json
frankm1870: The `tsBuildInfoFile` setting can't be set if `incremental` or `composite` are set, since it won't generate a `tsBuildInfoFile` without one of those being set.

src/client/scripts/esm/chess/variants/variant.ts
frankm1870: JsDoc does have a way to do typedefs like this; should I replace the function with a JsDoc typedef?

src/client/scripts/esm/chess/variants/variant.ts
frankm1870: While JsDoc can define types, it apparently can't define types with variable keys (where values are accessed with index notation instead of dot notation).

src/client/scripts/esm/chess/variants/variant.ts
frankm1870: The ts-ignore lines are required since the .js files don't have corresponding .d.ts files; they should be removable once everything is converted though.

Naviary3: What if we tried adding these to tsconfig?
```js
    "allowJs": true,        // Allows JavaScript files to be imported
    "checkJs": false,       // Disables type checking for JavaScript files
    "esModuleInterop": true, // Enables compatibility with ES module syntax
    "skipLibCheck": true    // Skips type-checking for declaration files
```

Naviary2:
> defaultSlideLimitForOldVariants should be an optional argument, instead of defaulting to infinity, as movesets.getPieceDefaultMovesets() already has a default value for it when not specified.

Did you see this comment?

src/client/scripts/esm/chess/variants/variant.ts
Naviary2: Actually gamerules.js already exports the type definition by exporting the function. Why not use that?

Turns out the import path was incorrect anyway, so variant.ts wasn't even using the GameRules import. I don't know why it wasn't throwing any errors. I corrected the path, and while it still compiles, there are lots of errors now related to the GameRules type that need to be fixed.

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -55,21 +55,20 @@ const gameruleModificationsOfOmegaShowcasings = { promotionRanks: null, moveRule
...
-/*interface GameRules {
+interface GameRuleModifications {
 	promotionRanks?: (number | null)[] | null,
 	moveRule?: number | null,
 	turnOrder?: string[],
 	promotionsAllowed?: {
-		white: string[],
-		black: string[]
+		white?: string[],
+		black?: string[]
```
Naviary2, line 64: In the future, we may have more colors. Lets just define promotionsAllowed as `Object<string, string[]>`

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -359,7 +358,7 @@ function getGameRulesOfVariant({ Variant, UTCDate = timeutil.getCurrentUTCDate()
 	if (variantEntry.gameruleModifications?.hasOwnProperty(0)) { // Multiple UTC timestamps
 		gameruleModifications = getApplicableTimestampEntry(variantEntry.gameruleModifications, { UTCDate, UTCTime });
 	} else { // Just one gameruleModifications entry
-		gameruleModifications = variantEntry.gameruleModifications;
+		gameruleModifications = variantEntry.gameruleModifications as GameRules;
```
Naviary2, line 361: Is this a typo? Is it supposed to be GameRuleModifications?

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -382,12 +381,12 @@ function getGameRules(modifications: {
 	},
 	turnOrder?: string[],
 	promotionsAllowed?: {
-		white: string[],
-		black: string[]
+		white?: string[],
```
Naviary2, line 384: Does it let us avoid redundancy here by just declaring the parameter as `modifications: GameRuleModifications`, instead of typing out the whole definition again?

src/client/scripts/esm/chess/variants/variant.ts
frankm1870: I believe so, yes.

src/client/scripts/esm/chess/variants/variant.ts
frankm1870: No, since there's an extra position property needed in the parameter.

Naviary2: Hmm, we should probably make `position` a 2nd parameter now instead of merged with the GameRuleModifications. 🤔

src/client/scripts/esm/chess/variants/gamerules.js
```diff
@@ -85,7 +87,7 @@ function GameRules() {
      * An object containing arrays of types white and black can promote to, if it's legal for them to promote.
      * If one color can't promote, their list should be left undefined.
      * If no color can promote, this should be left undefined.
-	 * @type {{ white: string[], black: string[] }}
+	 * @type {Object<string, string[]>}
```
Naviary2, line 90: `Object<string, string[]> | undefined`
as it's not required, and won't be present if `promotionRanks` is also undefined.

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -59,19 +59,19 @@ interface GameRuleModifications {
...
+	promotionsAllowed?: ColorVariantProperty<string[]>
+	winConditions?: ColorVariantProperty<string[]>
 	slideLimit?: number
 }
 
 type TimeVariantProperty<T> = T | {
 	[key: number]: T
 }
 
+type ColorVariantProperty<T> = {
```
Naviary2, line 71: Perhaps add comments to this and TimeVariantProperty of what their keys should be. For example, color, and a timestamp for the other, or an object of timestamps for the keys

PR #352, opened 2024-12-01
Add Huygens and ignore function.

lllllllllwith10ls:
Add the Huygens piece and code to implement more pieces like the Huygens.


src/client/scripts/esm/chess/logic/isprime.js
Naviary2, line 1: I would move this to esm/util, as its more so general to math

Naviary2:
I can do a full review later, but I want you to see the discussion on this
https://discord.com/channels/1114425729569017918/1312886799965749409

Naviary2:
Jace just converted movesets.js into typescript! It didn't change that much, but I'm letting you know since you will likely still have to copy over your moveset additions for the Huygen.

src/client/scripts/game/chess/movesets.js
Naviary2, line 1: This is the old location for the movesets.js file. It looks like your branch is still using this, you'll have to delete it and transfer over the HUYGEN moveset to esm/chess/logic/movesets.ts. You also *may* have to update the references to it too.

Naviary2:
This PR is waiting for updates and fixes. I'm leaving it open as it can still be used as inspiration if somebody else chooses to take this up.

Blocking functions have already been added, which allow you to use a custom function for what pieces block a specific piece (the default function if not provided is "If the piece is a friendly, block the square. Else if the piece is an enemy, block the square after it.").

What this PR mainly focuses on is adding Ignore functions, which are different from the blocking function, in that they do not block any squares, but rather they just tells you square exceptions that you cannot stop your slide on. For example, the Huygen (which is also in this PR) cannot stop its slide on composite numbers, only primes, and the royal queen cannot stop on squares that are in check.

Adding Ignore functions would give us the ability to add pieces with more complex movesets like the Huygen and royal queen, and it also unlocks a checkmate algorithm that would work in games with colinear lines (for example, a game with both a [1,0] slider and a [2,0] skipper, which lines are parallel and colinear.

PR #353, opened 2024-12-01
Add ignore function support

frankm1870:


src/client/scripts/esm/chess/logic/movesets.ts
```diff
@@ -9,28 +9,48 @@
...
- * @typedef {Object<string, PieceMoveset>} Movesets
  */
+type Movesets = {
+	[key: string]: PieceMoveset
+}
+
+/**
+ * The return value of the ignore function
+ */
+interface IgnoreResult {
```
Naviary2, line 24: So since the blocking function is now separate from the ignore function, I don't think we need this type anymore, as the ignore function and and blocking functions just return a Boolean.

src/client/scripts/esm/chess/logic/movesets.ts
```diff
@@ -9,28 +9,50 @@
...
+	slidingCheck: boolean
+}) => boolean) => IgnoreResult;
 
 /**
  * A moveset for an single piece type in a game
- * @typedef {Object} PieceMoveset
- * @property {number[][]} individual - Array of individual moves.
- * @property {Object<string, number[]>} [sliding] - Optional sliding moves.
  */
+interface PieceMoveset {
```
Naviary2, line 38: All of these are optional

frankm1870: Even individual moves? It looks like it should be required.

PR #360, opened 2024-12-05
Add Dockerfile and instructions on how to use it

joendter:
Having a Dockerfile that gives the project an uniform environment to run in should make it easier to do a minimal setup that **works**.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# adapted from https://www.docker.com/blog/getting-started-with-docker-using-node-jspart-i/
 FROM node:22-alpine
  
 WORKDIR /usr/app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:22-alpine
+ 
+WORKDIR /usr/app
+COPY ./ /usr/app
+RUN npm install
+ 
+CMD [ "npx", "nodemon" ]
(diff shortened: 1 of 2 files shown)


Naviary2:
I've never heard of Docker. Considering everyone would have to learn how to use it, and it may only reduce a couple steps in the setup process, I don't really see a need for this.

joendter:
It doesn't force anyone using docker, there are reasons to maybe not use it (the current implementation makes node reinstall all the packages every time you rebuild the docker image which takes its sweet time).
For those that want to use it and are already familiar it makes the setup process a decent bit more straightforward (for example that you don't need to worry that your package manager is shipping an outdated version of node).
It also ensures an uniform runtime environment and removes "works on my machine" errors. If the docker container works on one machine it works on any.

Also Docker is the industry standard for deploying web application (or actually *anything*).
https://www.docker.com/
https://aws.amazon.com/docker/

joendter:
But considering this would technically add more code to maintain (though the dockerfile only really needs to be updated if something in the setup changes or you want to use a different version of node), I understand that you don't want to merge it if you don't think the benefits are valuable for this project.

Naviary2:
Reasons that makes me lean away from this:

- Only one person so far would have and *could* have utilized this during the setup process (you), and it's working already for you, right? We have a small team so I don't think it'll be used all that much, and it'll have to be maintained.
- It adds another file to the root directory, which we are trying to minimize.
- I do think the Guide should be as straightforward as it can be. Rather than giving people 2 different options for setting up the project.. Why should they have to choose what installation method is best for them? Just give them clear steps to follow and that we know works.

Thus far we haven't really had any "works on my machine" problems that couldn't be solved. Most of the time people just didn't follow the Guide correctly, or had an old version of Node.js installed.

PR #362, opened 2024-12-07
5D variant now fills all boards with the standard position, and added…

frankm1870:
… an interface for a position.


Naviary2:
Looks good!

PR #366, opened 2024-12-09
Add an admin console

frankm1870:


src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+interface CustomRequest extends Request {
+	memberInfo: MemberInfo
+}
+
+type MemberInfo = {
+	signedIn: true,
+	user_id: number,
+	username: string,
+	roles: string[]
```
Naviary2, line 11: This should be `string[] | null`

src/server/controllers/roles.js
```diff
@@ -6,7 +6,7 @@
 import { logEvents } from "../middleware/logEvents.js";
 import { getMemberDataByCriteria, updateMemberColumns } from "../database/memberManager.js";
 
-const validRoles = ['owner','patron'];
+const validRoles = ['owner','patron','admin'];
```
Naviary2, line 9: Shift the admin role to the left of patron, as it will be higher priority

src/server/routes/root.js
```diff
@@ -14,8 +14,8 @@ const htmlDirectory = path.join(__dirname, "../../../dist/client/views");
  * @param {boolean} [isError=false] - If the file is an error page.
```
Naviary2, line 14: Can you update this parameter to be about the localized argument? Sorry I realized I have it describing an old unused argument.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
+import type { CustomRequest } from "../../types.js";
+import type { Response } from "express";
+// @ts-ignore
+import { deleteUser, getMemberDataByCriteria, updateMemberColumns } from "../database/memberManager.js";
+// @ts-ignore
+import { deleteRefreshTokensOfUser } from "../database/refreshTokenManager.js";
+// @ts-ignore
+import { logEvents } from "../middleware/logEvents.js";
+
+async function processCommand(req: CustomRequest, res: Response): Promise<void> {
```
Naviary2, line 10: Why does this need to be async?

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+	if (!req.memberInfo.signedIn) {
+		res.status(401).send("Cannot send commands while logged out.");
+		return;
+	}
+	if (!(req.memberInfo.roles?.includes("admin") ?? false)) {
+		res.status(403).send("Cannot send commands without the admin role");
+		return;
+	}
+	// TODO prevent affecting accounts with equal or higher roles
+	switch (commandAndArgs[0]) {
```
Naviary2, line 22: Can you have each of these case statements call an external function that handles the logic for their respective command? That'll shrink the switch statement by a ton.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+		case "ban":
+			return;
+		case "delete":
+			if (commandAndArgs.length < 2) {
+				res.status(400).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+				return;
+			}
+			deleteUser(getMemberDataByCriteria(["user_id"], "username", commandAndArgs[1])["user_id"], commandAndArgs[2] ?? "");
+			res.status(200).send("");
+			logEvents("Command executed: " + command + "\n", "adminCommands");
```
Naviary2, line 32: logEvents() automatically appends new lines.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+			return;
+		case "username":
+			if (commandAndArgs[1] === "get") {
+				if (commandAndArgs.length < 3) {
+					res.status(400).send("Invalid number of arguments, expected 2, got " + (commandAndArgs.length - 1) + ".");
+					return;
+				}
+				res.status(200).send(getMemberDataByCriteria(["username"], "user_id", Number.parseInt(commandAndArgs[2]!))["username"]);
+				logEvents("Command executed: " + command + "\nResult: " + getMemberDataByCriteria(["username"], "user_id", Number.parseInt(commandAndArgs[2]!))["username"] + "\n", "adminCommands");
```
Naviary2, line 41: These 2 lines:

If getMemberDataByCriteria() will return an empty object if the user is not found. If that happens it looks like it just returns that empty object to the user, which is fine.. but not sure if that's the behavior you wanted.

Also it should be called with the `skipErrorLogging` parameter true so that that function itself doesn't log an error if they don't exist.

"Number.parseInt(commandAndArgs[2]!)" is hard to read when it's inline inside the function. Can you pull it out into its own variable?

src/server/api/AdminPanel.ts
frankm1870: Because it contains an await statement.

src/server/api/AdminPanel.ts
frankm1870: I want a double newline between entries, since an entry may span multiple lines.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+			else {
+				res.status(400).send("Invalid subcommand, expected either get or set, got " + commandAndArgs[1] + ".");
+			}
+			return;
+		case "logout":
+			if (commandAndArgs.length < 2) {
+				res.status(400).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+				return;
+			}
+			deleteRefreshTokensOfUser(getMemberDataByCriteria(["user_id"], "username", commandAndArgs[1])["user_id"]);
```
Naviary2, line 60: Use sessionManager.deleteAllSessionsOfUser() instead. It makes it more clear what our intention is.

And again, if getMemberDataByCriteria() returns an empty object, skip calling this.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+			res.status(200).send("");
+			logEvents("Command executed: " + command + "\n", "adminCommands");
```
Naviary2, line 62: I'm noticing some redundancy with this line repeated. I'd stick this line in an external function `logCommandExecuted(command)` to avoid that.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+			return;
+		case "verify":
+			return;
+		case "post":
+			return;
+		case "invites":
+			return;
+		case "announce":
```
Naviary2, line 70: announce is uneeded, as toggling invite creation off will automatically announce.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+			return;
+		case "userinfo":
+			if (commandAndArgs.length < 2) {
+				res.status(400).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+				return;
+			}
+			res.status(200).send(getMemberDataByCriteria(["user_id", "username", "roles", "joined", "last_seen", "preferences", "verification", "username_history"],
```
Naviary2, line 77: ~~Again asking if you planned for the scenario where getMemberDataByCriteria() returns an empty object~~

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+				"username",
+				commandAndArgs[1]));
+			logEvents("Command executed: " + command + "\n" + getMemberDataByCriteria(["user_id", "username", "roles", "joined", "last_seen", "preferences", "verification", "username_history"],
```
Naviary2, line 80: The same exact getMemberDataByCriteria() call is 3 lines above. Perhaps place the results of it in a variable to avoid repeated called? It does perform database queries after all, which take a tiny bit of time, as it has to read the disk.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+				"username",
+				commandAndArgs[1]) + "\n", "adminCommands");
+			return;
+		case "help":
+			if (commandAndArgs.length === 1) {
+				res.status(200).send("Commands: ban, delete, username, logout, verify, post, invites, announce, userinfo, help\nUse help <command> to get more information about a command.");
```
Naviary2, line 86: It'd be good to place a string[] variable at the top of the file containing all the valid commands. Then here use validCommands.join(", ")

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+				return;
+			}
+			switch (commandAndArgs[1]) {
+				case "ban":
+					res.status(200).send("Syntax: ban email|ip|browser <username> <reason> [days]\nBans a user for a duration for the given reason.");
```
Naviary2, line 91: I think the ban command is just going to accept a username argument, and auto-ban their email, ip, and browser from that.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+					return;
+				case "unban":
+					res.status(200).send("Syntax: unban <username>\nUnbans the given user.");
```
Naviary2, line 94: And then unban will accept an email argument instead of username.
~~I've yet to figure out what to do to unban ips or browsers~~

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,131 @@
...
+					return;
+				case "delete":
+					res.status(200).send("Syntax: delete <username> [reason]\nDeletes the given user's account for an optional reason.");
+					return;
+				case "username":
+					res.status(200).send("Syntax: username get <userid>\n        username set <userid> <newUsername>\nGets or sets the username of the account with the given userid");
+					return;
+				case "logout":
+					res.status(200).send("Syntax: logout <username>\nLogs out the account with the given username.");
```
Naviary2, line 103: Maybe add "Logs out **all** sessions.", as there could be multiple.

src/server/api/AdminPanel.ts
frankm1870: Is a function with just one line really worth it though? It seems readable enough.

src/server/api/AdminPanel.ts
frankm1870: It could be useful for other announcements, e.g. if the server needs to go into maintenance or something else happens.

src/server/api/AdminPanel.ts
Naviary2: Nothing pops up when I searchfor "await".. 🤔 Perhaps there used to be but now it's gone?

src/server/api/AdminPanel.ts
Naviary2: Okay, that's fine. I wasn't sure if it was intended or not.

src/server/api/AdminPanel.ts
Naviary2: I guess that's true 🤔 It could broadcast one-time messages to everyone subscribed to a game or the invites list.

src/server/api/AdminPanel.ts
frankm1870: Oh that was the wrong function, I was thinking of `sendCommand` which does have an await statement.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function deleteCommand(command: string, commandAndArgs: string[], res: Response) {
+	logCommand(command);
+	if (commandAndArgs.length < 2) {
+		res.status(422).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+		return;
+	}
+	try {
+		deleteAccount(getMemberDataByCriteria(["user_id"], "username", commandAndArgs[1])["user_id"] ?? (() => { throw new Error(); })(), commandAndArgs[2] ?? "");
+	}
```
Naviary2, line 79: This lines really confusing. Can you space it all out? Something like:
```js
const username = commandAndArgs[1];
const reason = commandAndArgs[2];
if (!username || !reason) res.status(422).send("All parameters must be provided.");
const { user_id } = getMemberDataByCriteria(["user_id"], "username", username);
if (!user_id) res.status(422).send("User does not exist");
const success = deleteAccount(user_id, reason);
if (success) res.status(200).send("User deleted successfully");
else res.status(500).send("An error occurred while attempting to delete user.");
```
Then we can also skip the try-catch block altogether.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function usernameCommand(command: string, commandAndArgs: string[], res: Response) {
+	if (commandAndArgs[1] === "get") {
+		if (commandAndArgs.length < 3) {
+			logCommand(command);
+			res.status(422).send("Invalid number of arguments, expected 2, got " + (commandAndArgs.length - 1) + ".");
+			return;
+		}
+		const parsedId = Number.parseInt(commandAndArgs[2]!);
```
Naviary2, line 94: If this is NaN, we need to return an error, not moving on to call getMemberDataByCriteria().

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+		logEvents("Command executed: " + command + "\nResult: " + getMemberDataByCriteria(["username"], "user_id", parsedId)["username"] + "\n", "adminCommands");
+		const username = getMemberDataByCriteria(["username"], "user_id", parsedId)["username"];
```
Naviary2, line 96: Duplicate call of getMemberDataByCriteria() in the above 2 lines. Merge their result into one variable.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+		res.status(username === undefined ? 422 : 200).send(username ?? "User with id " + parsedId + " does not exist.");
```
Naviary2, line 97: I'd prefer if this was 2 separate send() calls
```js
if (username === undefined) res.status(422).send("User with id " + parsedId + " does not exist.");
else res.status(200).send(username);
```

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+			return;
+		default:
+			res.status(422).send("Unknown command.");
+			return;
+	}
+}
+
+function deleteCommand(command: string, commandAndArgs: string[], res: Response) {
+	logCommand(command);
+	if (commandAndArgs.length < 2) {
```
Naviary2, line 73: If we need a reason as well, should this be < 3?

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function logoutUser(command: string, commandAndArgs: string[], res: Response) {
+	logCommand(command);
+	if (commandAndArgs.length < 2) {
+		res.status(422).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+		return;
+	}
+	try {
+		deleteAllSessionsOfUser(getMemberDataByCriteria(["user_id"], "username", commandAndArgs[1])["user_id"] ?? (() => { throw new Error(); })());
```
Naviary2, line 119: Place `commandAndArgs[1]` into a `username` variable.
Avoid the try catch block by setting the results of getMemberDataByCriteria() to a variable. Then if the user_id is undefined we can return an error that the user doesn't exist.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+	}
+	catch {
+		res.status(422).send("User " + commandAndArgs[1] + " does not exist.");
+	}
+	res.status(200).send("");
```
Naviary2, line 124: Perhaps a message to let the admin know the command was a success?

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function getUserInfo(command: string, commandAndArgs: string[], res: Response) {
+	if (commandAndArgs.length < 2) {
+		logCommand(command);
+		res.status(422).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+		return;
+	}
+	const memberInfo = getMemberDataByCriteria(["user_id", "username", "roles", "joined", "last_seen", "preferences", "verification", "username_history"],
+		"username",
+		commandAndArgs[1]);
```
Naviary2, line 135: Place `commandAndArgs[1]` into a `username` variable.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,187 @@
...
+	logEvents("Command executed: " + command + "\n" + memberInfo + "\n", "adminCommands");
+	res.status(Object.keys(memberInfo).length === 0 ? 422 : 200).send(Object.keys(memberInfo).length === 0 ? "User " + commandAndArgs[1] + " does not exist." : memberInfo);
```
Naviary2, line 137: Also split this into 2 separate send() calls
```js
const memberInfoIsEmpty = Object.keys(memberInfo).length === 0;
if (memberInfoIsEmpty) res.status(422).send("User does not exist.");
else res.status(200).send(JSON.stringify(memberInfo));

**Comment by Naviary2** [comment]
Other than above, mentioned in discord but copying here so it's not forgotten:

Only log commands to the adminCommands log if the command had correct syntax. If the command had correct syntax, but failed, we can still log those.

File: src/server/api/AdminPanel.ts [hunk 26]
```diff
@@ -0,0 +1,208 @@
...
+	"post",
+	"invites",
+	"announce",
+	"userinfo",
+	"help"
+];
+
+function processCommand(req: CustomRequest, res: Response): void {
+	const command = req.params["command"]!;
+	const commandAndArgs = command.split(" ");
```
Naviary3, line 29: Use the chatgpt-provided function that splits the command string with a regex, allowing us to put parameters inside quotes (e.g. `"user request"`)

src/server/api/AdminPanel.ts
Naviary3: I did make it required in the database, it will not delete them without a valid reason.

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,208 @@
...
+			helpCommand(commandAndArgs, res);
+			return;
+		default:
+			res.status(422).send("Unknown command.");
+			return;
+	}
+}
+
+function deleteCommand(command: string, commandAndArgs: string[], res: Response) {
+	logCommand(command);
```
Naviary3, line 89: Move this call below the return if the number of parameters was invalid. (We should only log the command if the syntax is valid)

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,208 @@
...
+	if (commandAndArgs.length < 2) {
+		res.status(422).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+		return;
+	}
+	const reason = commandAndArgs[2];
+	const username = commandAndArgs[1];
+	const userid = getMemberDataByCriteria(["user_id"], "username", username)["user_id"];
+	if (userid !== undefined) {
+		deleteAccount(userid, reason ?? "");
```
Naviary3, line 98: deleteAccount() returns a boolean whether the deletion was a success or not. It could be confusing to the admin if they get a success message when it actually fails. If this is false, send the admin a failure message instead (the function does not return why it failed).

src/server/api/AdminPanel.ts
```diff
@@ -0,0 +1,208 @@
...
+function getUserInfo(command: string, commandAndArgs: string[], res: Response) {
+	if (commandAndArgs.length < 2) {
+		res.status(422).send("Invalid number of arguments, expected 1, got " + (commandAndArgs.length - 1) + ".");
+		return;
+	}
+	const username = commandAndArgs[1];
+	const memberInfo = getMemberDataByCriteria(["user_id", "username", "roles", "joined", "last_seen", "preferences", "verification", "username_history"],
+		"username",
+		username);
+	logEvents("Command executed: " + command + "\n" + memberInfo + "\n", "adminCommands");
```
Naviary3, line 169: Perhaps add "Result: " to the second line to be consistent when logging commands that include a result?

src/server/api/AdminPanel.ts
```diff
@@ -26,7 +24,26 @@ const validCommands = [
 
 function processCommand(req: CustomRequest, res: Response): void {
 	const command = req.params["command"]!;
-	const commandAndArgs = command.split(" ");
+	const commandAndArgs: string[] = [];
+	let inQuote: boolean = false;
+	let temp: string = "";
+	for (let i = 0; i < command.length; i++) {
```
Naviary3, line 30: This logic is great! Could you just move it to an external function, named `getArgumentsFromCommand()` or something?

PR #370, opened 2024-12-11
Remove client functions from movepiece.js

Idonotus:
Changed movepiece so that it does not animate or update the mesh of a move directly and instead creates a list of changes so that the gui can be changed accordingly

## Todo
- [x] State system
- [x] Clean up types
- [x] Test extreme desyncs
- [x] Document code
- [x] Clean up


Naviary3:
Let me know when you want a review of this!

src/client/scripts/esm/util/math.js
```diff
@@ -472,6 +472,16 @@ function roundUpToPowerOf2(num) {
 	return Math.pow(2, Math.ceil(Math.log2(num)));
 }
 
+/**
+ * 
+ * @param {Number} s 
+ * @param {Number} e 
+ * @param {Number} progress 
+ */
+function moveTowards(s, e, progress) {
```
Naviary2, line 481: Could you add a description for the method and each of the arguments? It's not clear what this does.

Naviary2:
Thanks! I did a functionality test and bug search! This post contains what I found!
https://discord.com/channels/1114425729569017918/1303279777452527636/1320338365786226688

Naviary2:
In general could you go through and add jsdoc comments to all the new interfaces, and methods, in boardchanges.ts, graphicalchanges.ts, and movesequence.ts? It all looks very well structured and organized, it would just help me pick up quicker how it all works!

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -0,0 +1,294 @@
...
+
+/**
+ * Apply changes in changelist according to changefuncs
+ * @param gamefile the gamefile
+ * @param changes the changes to apply
+ * @param funcs the object contain change funcs
+ */
+function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionList<genericChangeFunc>) {
+	for (const c of changes) {
+		if (!(c.action in funcs)) continue;
```
Naviary3, line 129: Would this be a source of a bug if the change type has a typo and we don't have a change function for it? Is it better to throw an error if the change type is unrecognized?

src/client/scripts/esm/chess/logic/movepiece.js
```diff
@@ -452,123 +308,84 @@ function rewindGameToIndex(gamefile, moveIndex, { removeMove = true, updateData
...
-	return info; // Info from simulating the move: { isCheck, gameConclusion }
+/**
+ * Simulates a move to get the check
+ * @param {gamefile} gamefile 
+ * @param {Move} move 
+ * @param {*} colorToTestInCheck 
+ * @returns 
+ */
+function getSimulatedCheck(gamefile, move, colorToTestInCheck) {
+	return simulateMoveWrapper(
```
Naviary3, line 356: It doesn't look like the simulated move wrapper knows when to perform game over checks or not, it always does by default. This is inefficient to run the checkmate algorithm when all we want to know is if the position is in check

src/client/scripts/esm/chess/logic/movepiece.js
```diff
@@ -452,123 +308,84 @@ function rewindGameToIndex(gamefile, moveIndex, { removeMove = true, updateData
...
+ * @param {gamefile} gamefile 
+ * @param {Move} move 
+ * @returns the gameConclusion
  */
-function stripSpecialMoveTagsFromCoords(coords) { return [coords[0], coords[1]]; }
+function getSimulatedConclusion(gamefile, move) {
+	return simulateMoveWrapper(
+		gamefile,
+		move,
+		() => wincondition.getGameConclusion(gamefile)
```
Naviary3, line 373: If makeMove() does game over checks by default, isn't this a repeated call to do the game over checks? Could we just check the gameConclusion on the gamefile?

src/client/scripts/esm/chess/logic/boardchanges.ts
Idonotus: If you look at graphicalchanges, sometimes we don't want to do anything with a change. Also ts will error if it doesn't match any change name.

src/client/scripts/esm/chess/logic/movepiece.js
Idonotus: Wdym?

Idonotus: `movepiece.makeMove` doesn't update conclusion anymore

Naviary2: Oh okay, there's 2 makeMove()'s

src/client/scripts/esm/chess/logic/boardchanges.ts
Naviary2: If typescript will error if you create a change with an invalid action, then by that logic we should have zero changes with an invalid action. I'm just going to throw an error if this ever ocurrs, I mean it might happen if we try creating a change in a javascript file, not typescript, then ts wouldn't warn us the change action was invalid.

Naviary2:
Kind of annoying how we need to replace the `move` action with the `capture` action, if there is a capture, instead of having both. But I see your reasoning behind it.

Keep in mind in the future there may be special moves that "capture" more than one piece, for instance, atomic chess explodes all adjacent pieces next to the captured piece. We can't go about adding 9 "capture" actions to the move, that's just redundant info... We ***could*** add 1 "capture" action and 8 "delete" actions, but then the exploded pieces will immediately vanish instead of waiting until after the moved piece finishes its animation.

Eventually, the "capture" action should be distinct and separate from the "move" action, and animation.js would know that all "capture" animations should wait a short delay until after the "move" animation has finished.

Naviary2:
The way state.ts is currently doesn't sit well with me. Any state you create can modify ***any*** gamefile property whatsoever. This could be the cause of bugs if incorrect paths are passed, and it would allow destructive gamefile edits. Like what if a state change modifies another move? or modifies the piece list? Sure, we *wouldn't* create a state change like that, but it's also not exactly clear what paths are allowed right now.

I'd much rather limit ourselves to just a few different possible state change types, and each one gets its own unique code for applying the state. Something like what I've just drafted up:

```js
/** Contains the local and global gamefile state changes from a Move */
interface MoveState {
	// Local statechanges are unique to the position you're viewing, and are ALWAYS applied.
	local: Array<StateChange>,
	// Global statechanges are a property of the game as a whole, not unique to the move,
	// and are not applied when VIEWING a move.
	// However, they are applied only when we make a new move, or rewind a simulated one.
	global: Array<StateChange>,
}

/** A state change, local or global, that contains enough information to set the gamefile's
 * property whether the move is being rewound or replayed. */
type StateChange = {
	// All state change types that may have unique "current" and "future" values,
	// and whos paths are always constant.
	type: 'check' | 'attackers' | 'enpassant' | 'moverulestate',
	// The gamefile's property of this type BEFORE this move was made, used to restore them when the move is rewinded.
	current: any,
	// The gamefile's property of this type AFTER this move was made, used to restore them when the move is replayed.
	future: any
} | {
	// This state change has a unique path.
	// It's current and future properties are also always constant (true and undefined)
	type: 'specialrights',
	/** The coordinate of the piece whos special rights are affected by the move */
	coords: Coords
}

/** Applies the state of a move to the gamefile. */
function applyState(gamefile: gamefile, state: StateChange, forward: boolean) {

	// The 'specialrights' state change is the only one without unique 'current' and 'future' properties
	if (state.type === 'specialrights') {
		const coordsKey = coordutil.getKeyFromCoords(state.coords);
		if (forward) delete gamefile.specialRights[coordsKey];
		else gamefile.specialRights[coordsKey] = true;					   
		return;
	}

	// All remaining possible state change types have unique 'current' and 'future' properties...

	const newValue = forward ? state.future : state.current;
	switch (state.type) {
		case 'check':
			gamefile.inCheck = newValue;
			break;
		case 'attackers':
			gamefile.attackers = newValue;
			break;
		case 'enpassant':
			gamefile.enpassant = newValue;
			break;
		case 'moverulestate':
			gamefile.moveRuleState = newValue;
			break;
	}
}
```

Idonotus:
I definitely agree with the removal of capture change. Maybe we can have a flag on the delete change indicating the piece was taken in a capture.

Issue #328, opened 2024-11-19
Feature - Premoving


just-another-nerd:
A lot of chess sites have this feature and I feel that infinitechess could use it.
I'm willing to work on this.

BF5258:
https://github.com/BF5258/infinitechess.org/tree/premoves
Feedback and suggestions are welcome.

Naviary2:
@just-another-nerd Oh my, I can't believe we don't have an issue for pre-moves yet! Thank you! Yes, the game is about ripe for adding pre-moves because even now when it's not your turn in an online game, you can select your peace to view its available moves! All we need to do is *allow you to make the pre-moves*.

Requirements:

* You should be allowed to make an infinite number of pre-moves at one time.
* the behavior will be like chess.com's pre-moves where the starting and ending squares of pre-moves are highlighted, and the piece that was pre-removed is now rendered at its new location.
* when you select a piece to be pre-removed it's legal moves should display it's passing through any pieces. In other words, you can't predict what is legal moves are going to be so just render its slides out to infinity.
* For now, premoves will always be enabled. No setting to disable them.

Naviary2:
A settings dropdown in a separate PR could be called "Selection", and will have 2 toggles for both "Drag Pieces" and "Premoving", as those seem related.

Naviary2:
#589

Naviary2:
#589

this-is-not-available:
Since #617 got merged, this should now be closed?

PR #377, opened 2024-12-18
Dragging Pieces + Selection menu

BF5258:
Adds selection menu with toggle switches to disable dragging pieces and premoves.
Relates to #22 and #328
Don't merge until after #350

## To do
How do you want the menu to look?

### Possible ideas
- [x] slider animation
- [ ] sound effect when switch flicked


BF5258:
This is want the menu looks like currently:
image
Any suggestions?

Naviary2:
Very nice!

An animation would be nice, as most other websites have that.
Also, perhaps add a border? To make the edge stand out more? They feel kinda sharp.
Then I'd shrink the buttons a bit.

Is there a color difference right now between the on and off states? I think it would be cool if the on-state color matched your color theme! For inspiration:

Screenshot 2024-12-18 at 12 33 04 AM

BF5258:
Some themes make the switch hard to see but a border should fix it.

https://github.com/user-attachments/assets/a7880a9c-db11-482c-b58b-c89894f8413e

Naviary2:
That looks really good!

BF5258:
There is a thin line of white pixels between the colored area and the border. Is it due to rounding error? How can I fix it?

image

Naviary2:
It just looks like its an issue with using outline. Could we get it work work with a border instead?

Naviary2:
Screenshot 2024-12-20 at 2 45 18 AM
Screenshot 2024-12-20 at 2 45 28 AM
Perhaps a white button would look better for our light theme. Similar to these. The off-state background color could be a tad darker tho imo

Naviary2:
Reviewed! With my notes in this post https://discord.com/channels/1114425729569017918/1115358966642393190/1320665870359593051

PR #378, opened 2024-12-22
GitHub Repo contributor list API

Naviary2:


Naviary2:
Nice!! I didn't even think of adding a command to manually update it!

Could you install the ESLint extension? I'm getting lots of errors where a semicolon is missing. After you do, to fix them all, hover over one error, click Quick Fix then Fix all Semi problems.

What API key do I use to test it? And do we all need to use our own? Am I the only one who can get the contributor list from the infinitechess.org repo?

PR #391, opened 2025-01-15
Organizedlines refactor

Idonotus:
Convert organized lines to typescript and remove all gui functions from it.


src/client/scripts/esm/chess/logic/organizedlines.ts
```diff
@@ -1,20 +1,50 @@
...
+// @ts-ignore
+import type gamefile from './gamefile.js';
+import type { Coords, CoordsKey } from '../util/coordutil.js';
+import type { Piece } from './boardchanges.js';
+
+// Amount of extra undefined pieces to store with each type array!
+// These placeholders are utilized when pieces are added or pawns promote!
+const extraUndefineds = 5; // After this many promotions, need to add more undefineds and recalc the model!
+
+class PooledArray<T> extends Array<T|undefined> {
```
Naviary3, line 22: Overlaps with gamefileutility's `TypeList` type. That script has a few organized piece types, I wasn't sure where else to put them, since this script was still in javascript, but I do think it's best for our types to be defined in here.

Also, Only the pieces organized by type object has undefined placeholders for each array. I think the pieces organized by lines should get their own type.

And, since this uses a class, I want to benchmark this later to see if there's a noticeable difference when in performance loading very large games.

src/client/scripts/esm/chess/logic/organizedlines.ts
```diff
@@ -1,20 +1,50 @@
...
+		this.undefineds = [];
+	}
+
+	addUndefineds() {
+		const insertedIndex = this.push(undefined) - 1; // insertedIndex = New length - 1
+		this.undefineds.push(insertedIndex);
+	}
+}
+
+interface OrganizedPieces {
```
Naviary3, line 35: Overlaps with gamefileutility's `PiecesByType` type.

Idonotus: The main branch does not have gamefileutility converted to ts. Should this be merging with gameloader?

src/client/scripts/esm/chess/logic/organizedlines.ts
```diff
@@ -1,20 +1,50 @@
...
+	[piecetype: string]: PooledArray<Coords>
+}
 
+interface LineOrganizer {
+	[line: LineKey]: Array<Piece>
+}
+
+interface Position {
```
Naviary3, line 43: Overlaps with gamefileutility's `PiecesByKey` type.

src/client/scripts/esm/chess/logic/organizedlines.ts
Naviary3: > The main branch does not have gamefileutility converted to ts. Should this be merging with gameloader?

Oh sorry, I missed that >_< sometimes when I switch up what I'm working on I just forget on what branch I made specific changes. Yeah I'd merge it with game-loader then!

PR #420, opened 2025-02-11
Update en-US.toml

noob10293:
Grammar fixes


Naviary2:
Looks good!

Normally modifying the TOMLs requires us to indicate what changes we made inside changes.json, and update the version number at the top of the file, that way our translators can know how to update their language's TOML. But, these grammatical fixes are small enough, I don't think that matters since grammar is different in every language anyway.

PR #425, opened 2025-02-12
Rewrite organized pieces

Idonotus:
The first part of the gameformat refactor. Implements a new format for organized pieces and transitions to integer formats.
# Todo
- [x] Organized pieces
- [x] Integer Types
    - [x] update piecemodel and grapical changes
- [x] Wrestle with format converter
- [x] Update boardchanges


src/client/scripts/esm/chess/logic/organizedpieces.ts
```diff
@@ -0,0 +1,354 @@
...
+// Separated from CoordsKey so that it's clear this is meant for directions, not coordinates
+type Vec2Key = `${number},${number}`;
+
+/** A unique identifier for a single line of pieces. `C|X` */
+type LineKey = `${number}|${number}`
+
+interface TypeRanges {
+	[type: number]: {
+		start: number,
+		end: number,
```
Naviary3, line 36: Good to comment whether it's inclusive or exclusive

src/client/scripts/esm/game/chess/gameslot.ts
```diff
@@ -299,9 +287,6 @@ function unloadGame() {
 	
 	selection.disableEditMode();
 	specialrighthighlights.onGameClose();
-
-	// Stop listening for the event that regenerates the mesh when more undefineds are inserted.
-	clientEventDispatcher.removeListener('inserted-undefineds', regenModelOfType);
```
Naviary2, line 304: We should still piecemodels.removeListener() here. addListeners() is inside loadGraphical() but there is no counterpart here.

Idonotus: since events are now per gamefile. We can discard the gamefile without removing the listeners

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -190,63 +180,62 @@ function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionL
...
-
-	// Remove captured piece from organized piece lists
-	organizedlines.removeOrganizedPiece(gamefile, piece.coords);
+	const pieces = gamefile.ourPieces;
+	const typedata = pieces.typeRanges.get(change.piece.type);
+	if (typedata === undefined) throw Error(`Type: "${change.piece.type}" is not expected to be in the game`);
+	if (change.piece.index === -1) throw Error("Piece has not been allocated in organizedPieces");
+	const idx = change.piece.index! + typedata.start;
+	organizedpieces.removePieceFromSpace(idx, pieces);
+	pieces.typeRanges.get(change.piece.type)!.undefineds.push(idx);
```
Naviary3, line 238: Let's keep the undefineds array sorted. Use jsutil.addElementToOrganizedArray()

src/client/scripts/esm/chess/util/boardutil.ts
```diff
@@ -0,0 +1,237 @@
...
+
+		getCoordsOfTypeRange(o, royalCoordsList, range);
+	}, [color], typeutil.jumpingRoyals);
+
+	return royalCoordsList;
+}
+
+function getCoordsOfTypeRange(o: OrganizedPieces, coords: Coords[], range: TypeRange) {
+	for (let idx = range.start; idx < range.end; idx++) {
+		if (range.undefineds.includes(idx)) continue;
```
Naviary3, line 146: undefineds is sorted. We can use binary search to drastically decrease the number of comparisons needed in editor mode. Use jsutil.binarySearch()

src/client/scripts/esm/chess/util/boardutil.ts
```diff
@@ -0,0 +1,237 @@
...
+
+// Getting A Single Piece -------------------------------------------------------------------------------------------------
+
+function getCoordsFromIdx(o: OrganizedPieces, idx: number): Coords {
+	return [o.XPositions[idx]!, o.YPositions[idx]!];
+}
+
+function isIdxUndefinedPiece(o: OrganizedPieces, idx: number): boolean {
+	// TODO: Use binary search here instead of linear search
+	return o.typeRanges.get(o.types[idx]!)!.undefineds.includes(idx);
```
Naviary3, line 159: Use jsutil.binarySearch()

src/client/scripts/esm/chess/util/typeutil.ts
```diff
@@ -0,0 +1,177 @@
...
+			callback(buildType(r, players[i]!));
+		}
+	}
+}
+
+function invertType(type: number): number {
+	const [r, p] = splitType(type);
+	const newp = p === players.WHITE ? players.BLACK :
+				 p === players.BLACK ? players.WHITE :
+				 undefined;
```
Naviary3, line 124: Instead of returning undefined here, which may be the source of a bug when we introduce more player colors and try to invert one of their types, lets use an IIFE here to throw an error:
```js
(() => { throw Error(`Cannot invert type of player ${p}!`); })();
```

src/client/scripts/esm/chess/util/typeutil.ts
```diff
@@ -0,0 +1,177 @@
...
+	if (newp === undefined) return type;
+	return buildType(r, newp);
+}
+
+function invertPlayer(player: Player) {
+	return player === players.WHITE ? players.BLACK :
+				 player === players.BLACK ? players.WHITE :
+				 undefined;
```
Naviary3, line 132: Instead of returning undefined here, which may be the source of a bug when we introduce more player colors and try to invert one of their types, lets use an IIFE here to throw an error:
```js
(() => { throw Error(`Cannot invert type of player ${p}!`); })();
```

src/client/scripts/esm/chess/util/typeutil.ts
```diff
@@ -0,0 +1,177 @@
...
+	for (let i = players.length - 1; i >= 0; i--) {
+		for (const r of includePieces) {
+			callback(buildType(r, players[i]!));
+		}
+	}
+}
+
+function invertType(type: number): number {
+	const [r, p] = splitType(type);
+	const newp = p === players.WHITE ? players.BLACK :
```
Naviary3, line 122: There's some repeat logic here. Use invertPlayer() instead.

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -190,63 +180,62 @@ function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionL
...
+			if (!organizedpieces.isTypeATypeWereAppendingUndefineds(gamefile.gameRules.promotionsAllowed, change.piece.type)) throw Error(`Type: ${change.piece.type} is not expected to be added after initial position`);
+			const regenData = organizedpieces.regenerateLists(gamefile.ourPieces, gamefile.gameRules, gamefile.listExtras);
+			events.runEvent(gamefile.events, "regenerateLists", gamefile, regenData);
+		}
+
+		idx = typedata.undefineds.pop()!;
+		change.piece.index = idx - typedata.start;
+	} else {
+		idx = typedata.start + change.piece.index;
+		const ri = typedata.undefineds.indexOf(idx);
```
Naviary3, line 217: Use jsutil.binarySearch()

src/client/scripts/esm/chess/util/boardutil.ts
```diff
@@ -0,0 +1,237 @@
...
+import type { OrganizedPieces, TypeRange } from "../logic/organizedpieces.js";
+import type { Coords } from "./coordutil.js";
+import type { RawType, Player } from "./typeutil.js";
+
+interface Piece {
+	type: number,
+	coords: Coords,
+	/** Relative to the start of its type range.
+	 * To get the actual idx, add the starting point of the type range */
+	index: number,
```
Naviary3, line 15: It's also probably good to indicate that if the index is -1 and the piece is for a board 'add' change, then piece may be inserted at the first available undefined index.

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -190,63 +180,62 @@ function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionL
...
+	if (typedata === undefined) throw Error(`Type: "${change.piece.type}" is not expected to be in the game`);
+	let idx;
+	if (change.piece.index === -1) {
+		if (typedata.undefineds.length === 0) {
+			if (!organizedpieces.isTypeATypeWereAppendingUndefineds(gamefile.gameRules.promotionsAllowed, change.piece.type)) throw Error(`Type: ${change.piece.type} is not expected to be added after initial position`);
+			const regenData = organizedpieces.regenerateLists(gamefile.ourPieces, gamefile.gameRules, gamefile.listExtras);
+			events.runEvent(gamefile.events, "regenerateLists", gamefile, regenData);
+		}
+
+		idx = typedata.undefineds.pop()!;
```
Naviary3, line 213: Let's shift() instead. I think it's better to fill type ranges from left to right, rather than from both sides at once.

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -190,63 +180,62 @@ function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionL
...
+		change.piece.index = idx - typedata.start;
+	} else {
+		idx = typedata.start + change.piece.index;
+		const ri = typedata.undefineds.indexOf(idx);
+		if (ri === -1) throw Error(`Piece ${change.piece} attemped to overwrite an occupied index`);
+		typedata.undefineds.splice(ri);
+	}
+	pieces.XPositions[idx] = change.piece.coords[0];
+	pieces.YPositions[idx] = change.piece.coords[1];
```
Naviary3, line 222: Correct me if wrong.. but shouldn't we also set `pieces.types[idx] = change.piece.type;` here? That could be the source of well hidden bugs..

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -190,63 +180,62 @@ function applyChanges(gamefile: gamefile, changes: Array<Change>, funcs: ActionL
...
-	gamefileutility.deleteIndexFromPieceList(list, piece.index);
-
-	// Remove captured piece from organized piece lists
-	organizedlines.removeOrganizedPiece(gamefile, piece.coords);
+	const pieces = gamefile.ourPieces;
+	const typedata = pieces.typeRanges.get(change.piece.type);
+	if (typedata === undefined) throw Error(`Type: "${change.piece.type}" is not expected to be in the game`);
+	if (change.piece.index === -1) throw Error("Piece has not been allocated in organizedPieces");
+	const idx = change.piece.index! + typedata.start;
+	organizedpieces.removePieceFromSpace(idx, pieces);
```
Naviary3, line 237: Yeah, removePieceFromSpace() is *enough* here. But to help make future bugs easier to spot, lets also set its index in XPositions, YPosition, and types, to 0. That way a void on (0,0) stands out like a sore thumb if we somehow start rendering undefined pieces...

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -280,17 +268,15 @@ function movePiece(gamefile: gamefile, change: Change) {
 function returnPiece(gamefile: gamefile, change: Change) {
 	if (change.action !== 'move' && change.action !== 'capture') throw new Error(`returnPiece called with a non-move change: ${change.action}`);
 
-	const piece = change.piece;
-	const endCoords = change.endCoords;
+	const pieces = gamefile.ourPieces;
+	const idx = pieces.coords.get(coordutil.getKeyFromCoords(change.endCoords))!;
```
Naviary3, line 272: Offload the responsibility for getting an index from a pair of coordinates to boardutil, let's use getIdxFromCoords()

src/client/scripts/esm/chess/logic/boardchanges.ts
```diff
@@ -261,17 +250,16 @@ function deletePiece(gamefile: gamefile, change: Change) {
 function movePiece(gamefile: gamefile, change: Change) {
 	if (change.action !== 'move' && change.action !== 'capture') throw new Error(`movePiece called with a non-move change: ${change.action}`);
 
-	const piece = change.piece;
-	const endCoords = change.endCoords;
+	const pieces = gamefile.ourPieces;
+	const idx = pieces.coords.get(coordutil.getKeyFromCoords(change.piece.coords))!;
```
Naviary3, line 254: Doesn't change.piece.index store the index we need?

src/client/scripts/esm/chess/logic/organizedpieces.ts
```diff
@@ -0,0 +1,510 @@
...
+	/** Each number in this array is the index of the undefined in the large XYPositions arrays. This array is also sorted. */
+	undefineds: Array<number>
+}
+
+type TypeRanges = Map<number, TypeRange>
+
+interface OrganizedPieces {
+	XPositions: PositionArray
+	YPositions: PositionArray
+	types: Uint8Array // Range 0-255. There are 22 total types currently, potentially 4 unique players/players in a game ==> 88 posible types.
```
Naviary3, line 46: We need jsdoc here to indicate what number undefined placeholders will be. See an above comment. I think we should make them 0? Unless we wanted to dedicate a type number for just null/empty pieces.

src/client/scripts/esm/chess/logic/organizedpieces.ts
```diff
@@ -0,0 +1,510 @@
...
+	o.YPositions = newYpos;
+	o.types = newTypes;
+
+	return extraUndefinedsByType;
+}
+
+function areWeShortOnUndefineds(o: OrganizedPieces, gamerules: GameRules): boolean {
+	for (const [t, range] of o.typeRanges) {
+		if (!isTypeATypeWereAppendingUndefineds(gamerules.promotionsAllowed!, t)) return false;
+		if (range.undefineds.length === 0) return true;
```
Naviary3, line 173: This func can be deleted, it's not used. boardchanges now regenerates the organized pieces as soon as it encounters an add change and no undefineds remaining.

src/client/scripts/esm/chess/util/boardutil.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+/** A unique identifier for a single line of pieces. `C|X` */
+type LineKey = `${number}|${number}`
+
+// Counting Pieces ----------------------------------------------------------------------------------------------
+
+/**
+ * Counts the number of pieces in the gamefile. Doesn't count undefined placeholders.
+ * @param gamefile - The gamefile object containing piece data.
```
Naviary3, line 32: Much of the jsdoc in boardutil.ts still includes the gamefile as an argument, which is outdated.

src/client/scripts/esm/chess/logic/organizedpieces.ts
Idonotus: I feel that overriding the type may be harmful. What if we negate the type to indicate its undefined

src/client/scripts/esm/chess/util/colorutil.ts
```diff
@@ -9,142 +9,9 @@
 
 // Type Definitions -----------------------------------------------------------------------------
 
-
 /** A color in a length-4 array: `[r,g,b,a]` */
 type Color = [number,number,number,number];
 
-
-// Variables -----------------------------------------------------------------------------
```
Naviary3, line 17: This was mentioned in the discord, but we could move the Color type to math.ts and delete colorutil.ts now

src/client/scripts/esm/chess/logic/organizedpieces.ts
Naviary2: Thing is, we're using an unsigned type array, which can't store negative numbers.

Why would it be harmful? Isn't the piece essentially deleted? When we rewind the game, when undo'ing the move's changes, it's type should be re-added to the same place. My guess is it will create more bugs if we *don't* delete it. Like, if when adding a piece we don't remember to set it's type, it could adopt the type of the deleted piece.

Naviary2: Yes let's leave `types` alone for undefined placeholders. Okay so then this just needs:

> We need jsdoc here for XPositions, YPositions, types, to indicate what number undefined placeholders will be. See an above comment. I think we should make them 0?

src/client/scripts/esm/chess/util/typeutil.ts
Idonotus: I'm not sure how i feel about throwing an error. Most of the type we anticipate player not being able to be reverted and ts warns us beforehand.

Naviary2: In this scenario, ts won't warn us if we were to pass in players 3 or above. Those will exist when we add multiplayer. This bug might go under the nose. Throwing an error forces us to notice it and patch it on the spot, then deciding what we want to do regarding accidentally reverting higher player numbers.

src/client/scripts/esm/chess/logic/boardchanges.ts
Naviary2: Yeah so I just forgot how the type ranges work..

PR #426, opened 2025-02-12
Decoupled repetition algorithm from checkmate algorithm, converted th…

Naviary3:
…em both to typescript. Also change the Move changes and State changes to intersection types.


Naviary2:
61 files changed!! Don't know what happened with that glitch. I just manually copied the changes over

PR #430, opened 2025-02-14
Merge checkmate-practice branch into main

noob10293:
Merged checkmate-practice branch into main and updated the code to work with the new systems.


translation/en-US.toml
```diff
@@ -146,6 +146,7 @@ error = "ERROR"
 [play.main-menu]
 credits = "Credits"
 play = "Play"
+practice = "Practice"
```
Naviary2, line 149: Changes to the english TOML file made be recorded inside translation/changes.json. When you note the changes, fill out the "note" value with a list of keys added and what line they were added on. Leave the "changes" field empty (but required or it will crash). After that, you should increment the "version" number at the top of the english TOML file.

src/client/scripts/esm/game/gui/style.js
```diff
@@ -32,6 +32,56 @@ function reinstateClass(element, className) {
...
+ */
+function hideElement(element) {
+	addClass(element, "hidden");
+}
+
+/**
+ * Reveals the provided document element by **removing** the class with the property "display: none".
+ * @param {HTMLElement} element - The document element
+ */
+function revealElement(element) {
```
Naviary2, line 47: revealElement() and hideElement() were deleted a while ago. They essentially just add an extra step. It's so easy to add/remove a class. Delete these methods and use `elementName.classList.add(className)` and `elementName.classList.remove(className)` instead.

Naviary2:
Bug- While the engine is thinking you can select their king and move it if you're quick (should not be able to select the engine's pieces).

Naviary2:
Bug- If you start a local game, then leave and open the Practice Checkmate menu, all the svgs are too big.

However, regarding the svgs.. The html of the Practice menu page almost **doubles** the length of the whole html! For that reason I think we should use javascript to construct the html and append it to the doc when the Practice menu is opened. It should be well worth the bandwidth saved, it's around 40KB by itself.

Given a checkmate practice id, `2Q-1k`, we can use a regex to split its parts up into groups of piece types (You can find a regex in generateCheckmateStartingPosition()).

The "collated-strong" class should be applied to all pieces in the same group, except the first piece.

But, if there's is only one white group, and that group has exactly 2 pieces, then the "collated" class is instead applied to the 2nd piece in the group.

Naviary2:
Another bug. This checkmate should have the royal centaur for black's svg. Code "2AM-1rc"

Screenshot 2025-02-14 at 2 14 31 PM

PR #433, opened 2025-02-16
Checkmate practice new page

frankm1870:


src/client/scripts/esm/game/gui/guipractice.ts
```diff
@@ -104,7 +107,7 @@ function initListeners() {
 function closeListeners() {
 	element_practiceBack.removeEventListener('click', callback_practiceBack);
 	element_checkmatePractice.removeEventListener('click', callback_checkmatePractice);
-	element_tacticsPractice.removeEventListener('click', gui.displayStatus_FeaturePlanned);
+	element_tacticsPractice.removeEventListener('click', () => statustext.showStatus(translations['planned_feature']));
```
Naviary2, line 110: This will actually fail to remove the event listener added above, because they are 2 different functions in memory, even though they do identical things.

PR #439, opened 2025-02-17
Fix broken header on firefox

tsevasa:
Before, on firefox:
grafik

After, on firefox:
grafik


tsevasa:
Up to now, I tested this change on Edge and Chrome on my computer and it seemed fine.

PR #486, opened 2025-03-17
Fix formatconverter special rights bug & general overhaul

tsevasa:


tsevasa:
Also replaced `parseInt` by `Number` in one place

PR #507, opened 2025-04-01
Abridged gamefile retains enpassant accuracy for 4d games

Naviary3:


Naviary2:
Too hard to go back and forth between formatconverter having to convert the enpassant property to the gamefile's EnPassant property, and then back to a single square that the formatconverter to understand.

PR #510, opened 2025-04-02
Database upgrade: Add player_stats and games table to database

tsevasa:


src/server/database/databaseTables.js
```diff
@@ -77,6 +116,51 @@ function generateTables() {
 	`;
 	db.run(createTableSQLQuery);
 
+	// Player Stats table
+	createTableSQLQuery = `
+		CREATE TABLE IF NOT EXISTS player_stats (
+			user_id INTEGER PRIMARY KEY REFERENCES members(user_id) ON DELETE CASCADE,
+			last_played_rated_game TIMESTAMP,
+			game_history TEXT NOT NULL DEFAULT '',
```
Naviary2, line 124: Did you know queries can have comments by using `--`?

src/server/database/memberManager.js
```diff
@@ -133,7 +133,7 @@ function deleteUser(user_id, reason_deleted, { skipErrorLogging } = {}) {
 	} catch (error) {
 		// Log the error for debugging purposes
 		logEvents(`Error deleting user with ID "${user_id}": ${error.stack}`, 'errLog.txt', { print: true });
-		return { succes: false, reason: `Failed to delete user of ID "${user_id}", an internal error ocurred.`};
+		return { success: false, reason: `Failed to delete user of ID "${user_id}", an internal error ocurred.`};
```
Naviary2, line 136: What the.. how did I miss that lol

src/server/database/ratingsManager.ts
```diff
@@ -29,10 +29,10 @@ type ModifyQueryResult = { success: true; result: RunResult } | { success: false
 // Methods --------------------------------------------------------------------------------------------
 
 
-/*
+/**
  * Adds an entry to the ratings table, defaulting to 1000 elo and 350 rd (assuming DB defaults)
- * @param user_id - The id for the user (fails if it doesn't exist in members or due to constraints)
- * @returns A result object indicating success or failure.
+ * @param {number} user_id - The id for the user (fails if it doesn't exist in members or due to constraints)
+ * @returns {ModifyQueryResult} A result object indicating success or failure.
```
Naviary2, line 35: Since typescript files already include argument types and the return type in the function declaration line, I would prefer not having them declared a second time in the jsdoc..

src/server/database/gamesManager.ts
```diff
@@ -0,0 +1,263 @@
...
+
+// Methods --------------------------------------------------------------------------------------------
+
+/**
+ * Adds an entry to the games table
+ * @param {number} game_id - The id for the game
+ * @param {object} [options] - Optional parameters for all the entries of the game
+ * @returns {ModifyQueryResult} A result object indicating success or failure.
+ */
+function addGameToGamesTable(game_id: number,
```
Naviary2, line 49: I reckon game_id shouldn't be a parameter, it should be gamesManager's responsibility to pick an id for the game, one that isn't already taken. Use memberManager.genUniqueUserID() for inspiration.

src/server/database/gamesManager.ts
```diff
@@ -0,0 +1,263 @@
...
+		// Return an error message
+		// Check for specific constraint errors if possible (e.g., FOREIGN KEY failure)
+		let reason = 'Failed to add game to games table.';
+		if (error instanceof Error && 'code' in error) {
+			// Example check for better-sqlite3 specific error codes
+			if (error.code === 'SQLITE_CONSTRAINT_UNIQUE' || error.code === 'SQLITE_CONSTRAINT_PRIMARYKEY') {
+				reason = 'Game ID already exists in the games table.';
+			}
+		}
+		return { success: false, reason };
```
Naviary2, line 120: Just a thought (you can resolve after you've read this).. If for whatever reason gamesManager fails to add a game to the table, whatever script that **called** addGameToGamesTable() I think should instead log the game (or more simply, all of its custom arguments) to an ordinary log file, just so that the game isn't ENTIRELY lost..

src/server/database/gamesManager.ts
```diff
@@ -0,0 +1,263 @@
...
+	}
+}
+
+/**
+ * Fetches specified columns of a single game from the games table based on game_id
+ * @param {string[]} columns - The columns to retrieve (e.g., ['game_id', 'date', 'players']).
+ * @param {number} game_id - The game_id of the game
+ * @returns {GamesRecord} - An object containing the requested columns, or undefined if no match is found.
+ */
+function getGameData(columns: string[], game_id: number): GamesRecord | undefined {
```
Naviary2, line 130: Minor nitpick. Maybe we should put game_id first, since columns might be very long?

src/server/database/playerStatsManager.ts
```diff
@@ -0,0 +1,198 @@
...
+
+
+/**
+ * Adds an entry to the player_stats table
+ * @param {number} user_id - The id for the user (fails if it doesn't exist in player_stats or due to constraints)
+ * @param {object} [options] - Optional parameters for the user.
+ * @param {Date} [options.last_played_rated_game] - The user's last_played_rated_game timestamp
+ * @returns {ModifyQueryResult} A result object indicating success or failure.
+ */
+function addUserToPlayerStatsTable(user_id: number, options: { last_played_rated_game?: Date } = {}): ModifyQueryResult {
```
Naviary2, line 56: last_played_rated_game is the only column we can allow to be `null`. I don't think we need it as an argument here. As a user finishes games, and those games are added to the games table, the information in the player stats table will be updated.

src/server/database/playerStatsManager.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+}
+
+/**
+ * Fetches specified columns of a single player from the player_stats table based on user_id
+ * @param {string[]} columns - The columns to retrieve (e.g., ['user_id', 'moves_played', 'last_played_rated_game']).
+ * @param {number} user_id - The user_id of the player
+ * @returns {PlayerStatsRecord} - An object containing the requested columns, or undefined if no match is found.
+ */
+function getPlayerStatsData(columns: string[], user_id: number): PlayerStatsRecord | undefined {
```
Naviary2, line 97: Minor complaint: Maybe user_id should be the first argument, since columns may be very long?

src/server/database/playerStatsManager.ts
```diff
@@ -0,0 +1,198 @@
...
+import type { RunResult } from 'better-sqlite3'; // Import necessary types
+
+
+// Type Definitions -----------------------------------------------------------------------------------
+
+
+/** Structure of a player_stats record. This is all allowed columns of a user_id. */
+interface PlayerStatsRecord {
+	user_id?: number;
+    last_played_rated_game?: Date;
```
Naviary2, line 23: Crazily enough, SQLite does not store "TIMESTAMP" values as a Date, but rather a string with its own unique format. We can expect the value of the cell to be a string. To convert it to, say, a timestamp in milliseconds since the Unix epoch, we can call timeutil.sqliteToTimestamp(). (There are a couple other useful methods in that script)

src/server/database/gamesManager.ts
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Adds an entry to the games table
+ * @param {number} game_id - The id for the game
+ * @param {object} [options] - Optional parameters for all the entries of the game
+ * @returns {ModifyQueryResult} A result object indicating success or failure.
+ */
+function addGameToGamesTable(game_id: number, 
+	options: {
+        date?: Date,
```
Naviary2, line 51: In an previous comment, I explained the format SQLite stores dates as strings. I think this argument can be of the string type. This script shouldn't be responsible for making sure it's in the right format, but another script can use a timeutil method to convert a timestamp into SQLite's date format before calling addGameToGamesTable()

src/server/database/gamesManager.ts
```diff
@@ -0,0 +1,267 @@
...
+// Methods --------------------------------------------------------------------------------------------
+
+/**
+ * Adds an entry to the games table
+ * @param {number} game_id - The id for the game
+ * @param {object} [options] - Optional parameters for all the entries of the game
+ * @returns {ModifyQueryResult} A result object indicating success or failure.
+ */
+function addGameToGamesTable(game_id: number, 
+	options: {
```
Naviary2, line 50: None of these options have default values in the table, which means they all should be required.

src/server/database/databaseTables.js
tsevasa: No! Good to know, thanks!

PR #518, opened 2025-04-14
Create 2024-11-22 and translated it

cycy98:
Sorry for being late


cycy98:
It's not the only one, I'm gonna trnaslate more today
The translation is verified, so there should be no mistakes

PR #524, opened 2025-04-15
Add player_games database table, and related changes

tsevasa:


src/server/api/Member.js
```diff
@@ -23,13 +23,16 @@ const getMemberData = async(req, res) => { // route: /member/:member/data
...
-	// Get the player's rating values
-	const rating_values = getPlayerLeaderboardRating(user_id, Leaderboards.INFINITY); // { user_id, elo, rating_deviation, last_rated_game_date } | undefined
-	if (!rating_values) {
-		logEvents(`Error getting rating values for member "${claimedUsername}" on INFINITY leaderboard. Not found.`, 'errLog.txt', { print: true });
-		return res.status(500).send('Internal Server Error');
+	// Get the player's elo from the INFINITY leaderboard, IF user entry exists
+	let ranked_elo = "1000?";
+	if (isPlayerInLeaderboard(user_id, Leaderboards.INFINITY)) {
+		const rating_values = getPlayerLeaderboardRating(user_id, Leaderboards.INFINITY); // { user_id, elo, rating_deviation, last_rated_game_date } | undefined
+		if (!rating_values) {
```
Naviary3, line 30: leaderboardManager should store the default elo. What if here we instead called getPlayerRankInLeaderboard(), and modify that method so that it returns "1000?" if they aren't on that leaderboard?

src/server/database/gamesManager.ts
```diff
@@ -49,13 +47,11 @@ type ModifyGameQueryResult = { success: true; result: RunResult } | { success: f
 function addGameToGamesTable(
 	options: {
         date: string,
-        players: string,
-        elo: string | null,
-        rating_diff: string | null,
         time_control: string,
         variant: string,
 		/** 0 => false  1 => true */
         rated: number,
```
Naviary3, line 53: This wasn't you, but we could update `rated` now to of type 0 | 1, since the ADD table query enforces it be one of these two values

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -148,13 +131,44 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
+	}
+}
+
 /**
  * Update's each member's game stats in the player stats table according to the results of this game.
- * Such as game count, wins, losses, game_history, etc.
+ * Such as game count, wins, losses, etc.
  */
-function updatePlayerStatsTable(game: Game, game_id: number, dateSqliteString: string) {
+async function updatePlayerStatsTable(game: Game, game_id: number, victor: Player | undefined) {
```
Naviary3, line 170: Is there a reason updatePlayerStatsTable() needs to be async? We shouldn't have to await for logEvents() to finish.

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -148,13 +131,44 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
 	return ICN;
 }
 
+/**
+ * For each member, add an entry into player_games according to the results of this game.
+ * If the game was ranked, also update the leaderboards table accordingly.
+ */
+async function updatePlayerGamesTable(game: Game, game_id: number) {
```
Naviary3, line 138: Does this need to be async too? For same reason above.

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -148,13 +131,44 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
+	for (const playerStr in game.players) {
+		const player: Player = Number(playerStr) as Player;
+		const user_id = game.players[playerStr].identifier.user_id;
+		if (user_id === undefined) continue; // Guest players don't get an entry in the player_games table or an elo for updating
+
+		const elo_at_game = null;
```
Naviary3, line 144: We can at least specify elo_at_game, even if the game wasn't rated. It still could be interesting to know their elo at the time of the game.

elo_change_from_game we keep null unless a game was rated

src/server/api/Member.js
tsevasa: Good idea, done!

src/server/database/leaderboardsManager.ts
```diff
@@ -310,6 +313,25 @@ function getPlayerRankInLeaderboard(user_id: number, leaderboard_id: Leaderboard
...
+/**
+ * Gets a string containing the display value for the rating of a player on a specific leaderboard.
+ * @param user_id - The id for the user
+ * @param leaderboard_id - The id for the specific leaderboard.
+ * @returns The player's leaderboard display string
+ */
+function getDisplayEloOfPlayerInLeaderboard(user_id: number, leaderboard_id: Leaderboard): string {
+	let ranked_elo = `${String(DEFAULT_LEADERBOARD_ELO)}?`;
+	if (isPlayerInLeaderboard(user_id, leaderboard_id)) {
+		const rating_values = getPlayerLeaderboardRating(user_id, leaderboard_id); // { user_id, elo, rating_deviation, last_rated_game_date } | undefined
```
Naviary3, line 325: Nice addition of getDisplayEloOfPlayerInLeaderboard()!

isPlayerInLeaderboard() is probably an unnecessary step here though, getPlayerLeaderboardRating() already returns undefined if they aren't on the leaderboard, we can assume it is correct.

src/server/database/playerGamesManager.ts
```diff
@@ -0,0 +1,201 @@
...
+		// Log the error and return undefined
+		logEvents(`Error executing query when gettings player game of user_id ${user_id} and game_id = ${game_id}: ${message}. The query: "${query}"`, 'errLog.txt', { print: true });
+		return undefined;
+	}
+}
+
+/**
+ * Updates multiple column values in the player_games table for a given user.
+ * @param user_id - The user ID of the player_games.
+ * @param game_id - The game_id of the game
```
Naviary3, line 142: We should probably add emphasis in the jsdoc this function should never have to be called, as all information should be correct when we INSERT rows into this table.

src/server/game/gamemanager/gamelogger.ts
tsevasa: Yes, that is an interesting thought, let's go with that and take the elo from the leaderboard_id corresponding to the played variant, if applicable.

src/server/database/leaderboardsManager.ts
tsevasa: Good point

src/server/database/playerGamesManager.ts
```diff
@@ -0,0 +1,201 @@
+/**
+ * This script handles queries to the player_games table. 
+ */
```
Naviary3, line 3: I think playerGamesManager could use a method, getPlayersInGame(), that will find all players that participated in a game of given game_id, and return them all as a `PlayerGamesRecord[]`

PR #525, opened 2025-04-15
Formatconverter upgrade: better parsing for metadata and gamerules

tsevasa:


src/client/scripts/esm/chess/logic/formatconverter.js
```diff
@@ -493,6 +463,54 @@ function ShortToLong_Format(shortformat/*, reconstruct_optional_move_flags = tru
...
 
+		// metadata key is read: enter metadata parsing mode, if string starts with [ and letter
+		if (!in_metadata_parsing_mode && !in_gamerules_parsing_mode && /^\[[a-zA-Z]/.test(string)) {
+			in_metadata_parsing_mode = true;
+			metadata_key = string.slice(1);
+			continue;
+		}
+
+		// Read metadata value, if in metadata parsing mode
+		if (in_metadata_parsing_mode && !in_gamerules_parsing_mode) {
```
Naviary3, line 474: in_metadata_parsing_mode shouldn't be true if in_gamerules_parsing_mode is on, right? Can we simplify?
```js
if (in_metadata_parsing_mode) {
```

src/client/scripts/esm/chess/logic/formatconverter.js
```diff
@@ -493,6 +463,54 @@ function ShortToLong_Format(shortformat/*, reconstruct_optional_move_flags = tru
...
+			// remove " from the start of string if possible
+			if (/^"/.test(string) && metadata_value === "") {
```
Naviary3, line 476: Does this still metadata parser still work for metadata in the form `[MetaData "value"]`? With all logged games converted to the newest format I don't think we should support backwards compatibility anymore.

src/client/scripts/esm/chess/logic/formatconverter.js
```diff
@@ -493,6 +463,54 @@ function ShortToLong_Format(shortformat/*, reconstruct_optional_move_flags = tru
...
+
+		// gamerules - start: read in string and enter gamerules parsing mode, if string starts with {
+		if (!in_metadata_parsing_mode && !in_gamerules_parsing_mode && /^\{/.test(string) && gamerules_string === "") {
+			in_gamerules_parsing_mode = true;
+			gamerules_string = string;
+			string = ""; // this line is used instead of continue; so that we immediately enter the gamerules continuation below and check if isJson(gamerules_string)
+		}
+
+		// Read gamerules continuation, if in gamerules parsing mode
+		if (!in_metadata_parsing_mode && in_gamerules_parsing_mode) {
```
Naviary3, line 502: We could simplify since we know we can't be in both parsing modes at once
```js
if (in_gamerules_parsing_mode) {
```

tsevasa: Yes, see above

PR #546, opened 2025-05-14
Add rated games

tsevasa:


src/client/scripts/esm/chess/variants/leaderboard.ts
```diff
@@ -0,0 +1,70 @@
...
+const UNCERTAIN_LEADERBOARD_RD = 250.0;
+
+/** Constant c, used for Glicko-1 */
+const GLICKO_ONE_C = 70;
+
+/** Constant q, used for Glicko-1 */
+const GLICKO_ONE_Q = 0.00575646273;
+
+/** Duration of a glicko-1 rating period, in milliseconds */
+const RATING_PERIOD_DURATION = 1000 * 60 * 60 * 24 * 15; // 15 days
```
Naviary2, line 28: Maybe these constants could be moved to ratingcalculation? The client doesn't need them.

translation/en-US.toml
```diff
@@ -119,6 +119,7 @@ joined = "Joined:"
 seen = ["Seen:", " ago"]
 practice_progress = "Practice Mode Progress:"
 ranked_elo = "Elo:"
+infinity_leaderboard_position = "Infinity Leaderboard Position:"
```
Naviary2, line 122: What about "Global Ranking"? The users may not know it's the main leaderboard.

src/server/database/databaseTables.js
```diff
@@ -117,7 +117,7 @@ function generateTables() {
 		CREATE TABLE IF NOT EXISTS leaderboards (
         	user_id INTEGER NOT NULL REFERENCES members(user_id) ON DELETE CASCADE,
    			leaderboard_id INTEGER NOT NULL, -- Each leaderboard's id and variants are declared in the code
-			elo REAL NOT NULL DEFAULT 1000.0,
+			elo REAL NOT NULL DEFAULT 1500.0,
```
Naviary2, line 120: Let's reference the DEFAULT_LEADERBOARD_ELO constant here

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -136,21 +139,74 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
+
+	const leaderboard_id = VariantLeaderboards[game.variant];
+	if (leaderboard_id === undefined) return {}; // If game belongs to no valid leaderboard_id, then no ratings get updated
+
+	// Loop over all players of game in order to construct ratingdata object
+	let ratingdata : RatingData = {};
+	for (const playerStr in game.players) {
+		const player: Player = Number(playerStr) as Player;
+		const user_id = game.players[playerStr].identifier.user_id;
+		if (user_id === undefined) return {}; // If a player doesn't exist, then no ratings get updated
```
Naviary2, line 155: If we reached this point when game.rated is true, it's a bug. I think we should log it in the errLog before returning.

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -136,21 +139,74 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
+
+		// If player is not on leaderboard, add him to it
+		if (!isPlayerInLeaderboard(user_id, leaderboard_id)) addUserToLeaderboard(user_id, leaderboard_id);
+
+		// Access the player leaderboard data
+		const leaderboard_data = getPlayerLeaderboardRating(user_id, leaderboard_id);
+		if (leaderboard_data === undefined || leaderboard_data?.elo === undefined || leaderboard_data?.rating_deviation === undefined) {
+			console.log(`Unable to correctly process leaderboard_data of user ${user_id} and leaderboard ${leaderboard_id}.`);
```
Naviary2, line 163: Good log, but let's take it a step further and log it to the errLog

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -136,21 +139,74 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
+	ratingdata = computeRatingDataChanges(ratingdata, victor);
+
+	// Update the rating data of each player in leaderboard table
+	for (const playerStr in game.players) {
+		const player: Player = Number(playerStr) as Player;
+		const user_id = game.players[playerStr].identifier.user_id;
+
+		const elo = ratingdata[player]?.elo_after_game;
+		const rd = ratingdata[player]?.rating_deviation_after_game;
+		if (elo === undefined || rd === undefined) continue;
```
Naviary2, line 184: ratingdata[player] we know will be defined. I think it'd be already to use non null assertion operators here `!`

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -136,21 +139,74 @@ async function getICNOfGame(game: Game, metadata: MetaData): Promise<string | un
...
 		if (user_id === undefined) continue; // Guest players don't get an entry in the player_games table or an elo for updating
 
 		const score = victor === undefined ? null : victor === player ? 1 : victor === players.NEUTRAL ? 0.5 : 0;
-
-		// TODO: Implement the following when rated games are here
-		// We can potentially get the elo_at_game entries from the rating calculation and not have to do it again here, if the game was rated
-		const elo_at_game = null;
-		const elo_change_from_game = null;
+		const elo_at_game = ratingdata[player]?.elo_at_game ?? null;
+		const elo_change_from_game = ratingdata[player]?.elo_change_from_game ?? null;
```
Naviary2, line 209: These are not allowed to be `NULL` in the database. It's alright though we already know there'll all present, so we could use the non null assertion operator again

src/server/socket/openSocket.ts
```diff
@@ -77,6 +79,13 @@ function onConnectionRequest(socket: WebSocket, req: IncomingMessage) {
 
 	addListenersToSocket(req, ws);
 
+	// If user is signed in, use the database to correctly set the property ws.metadata.verified
+	if (ws.metadata.memberInfo.signedIn && ws.metadata.memberInfo?.user_id !== undefined) {
+		let { verification } = getMemberDataByCriteria(['verification'], 'user_id', ws.metadata.memberInfo.user_id, { skipErrorLogging: true });
```
Naviary2, line 84: Maybe we could cast it to the `Verification` type in verifyAccountController so it' sclear what properties it has.

src/server/socket/openSocket.ts
```diff
@@ -77,6 +79,13 @@ function onConnectionRequest(socket: WebSocket, req: IncomingMessage) {
...
+		verification = JSON.parse(verification); // string needs to be parsed to a JSON
```
Naviary2, line 85: Maybe we could cast it to the `Verification` type in verifyAccountController so it's clear what properties it has.

src/server/controllers/verifyAccountController.ts
```diff
@@ -95,6 +95,9 @@ async function verifyAccount(req: CustomRequest, res: Response) {
 	// VERIFY THEM..
 	verification = getNewVerificationAfterVerifying();
 
+	// Informs all sockets of the user that he is now verified
+	AddVerificationToAllSocketsOfMember(user_id);
```
Naviary2, line 99: This same line should also be added below in manuallyVerifyUser()

src/client/scripts/esm/chess/variants/leaderboard.ts
tsevasa: Done!

src/server/game/gamemanager/gamelogger.ts
tsevasa: sure

PR #553, opened 2025-05-21
Split `legalmoves.calculate` into individual functions

Idonotus:


src/client/scripts/esm/chess/logic/movepiece.ts
```diff
@@ -430,9 +431,11 @@ function calculateMoveFromShortmove(gamefile: gamefile, shortmove: ServerGameMov
 		throw Error(`Failed to calculate Move from shortmove because there's no piece on the start coords: ${shortmove}`);
 	}
 
-	const legalSpecialMoves: Coords[] = legalmoves.calculate(gamefile, piece, { onlyCalcSpecials: true }).individual;
-	for (let i = 0; i < legalSpecialMoves.length; i++) {
-		const thisCoord: Coords = legalSpecialMoves[i]!;
+	const moveset = legalmoves.getPieceMoveset(gamefile, piece.type);
+	const legalSpecialMoves = legalmoves.getEmptyLegalMoves(moveset);
+	legalmoves.appendSpecialMoves(gamefile, piece, moveset, legalSpecialMoves);
+	checkresolver.removeCheckInvalidMoves(gamefile, piece, legalSpecialMoves);
```
Naviary3, line 437: I'm just considering optimizations here.. I don't think we need to remove check invalid moves here since we wouldn't have moved to the square if it was in check, we're just testing to see if the move matches any special move.

src/client/scripts/esm/game/chess/selection.ts
```diff
@@ -419,7 +419,11 @@ function initSelectedPieceInfo(gamefile: gamefile, piece: Piece) {
 	// Initiate
 	pieceSelected = piece;
 	// Calculate the legal moves it has. Keep a record of this so that when the mouse clicks we can easily test if that is a valid square.
-	legalMoves = legalmoves.calculate(gamefile, pieceSelected);
+	const moveset = legalmoves.getPieceMoveset(gamefile, pieceSelected.type);
+	legalMoves = legalmoves.getEmptyLegalMoves(moveset);
+	legalmoves.appendCalculatedMoves(gamefile, pieceSelected, moveset, legalMoves);
+	legalmoves.appendSpecialMoves(gamefile, pieceSelected, moveset, legalMoves);
+	checkresolver.removeCheckInvalidMoves(gamefile, pieceSelected, legalMoves);
```
Naviary3, line 426: These exact 5 lines are found at 3 places in the code. Would you be against us adding an additional method for calling them all automatically? Smth like this the following:

```js
function calculateAll(gamefile: gamefile, piece: Piece): LegalMoves {
    const moveset = legalmoves.getPieceMoveset(gamefile, piece.type);
    const moves = legalmoves.getEmptyLegalMoves(moveset);
    legalmoves.appendCalculatedMoves(gamefile, piece, moveset, moves);
    legalmoves.appendSpecialMoves(gamefile, piece, moveset, moves);
    checkresolver.removeCheckInvalidMoves(gamefile, piece, moves);
    return moves;
}
```

And we could retain each individual method, so that when we need more specific legal move calculation we can call each method individually. This way we reduce the newly-added redundancy.

src/client/scripts/esm/game/rendering/arrows/arrowlegalmovehighlights.ts
```diff
@@ -101,7 +102,12 @@ function onPieceIndicatorHover(piece: Piece) {
 	// Calculate their legal moves and mesh!
 	const gamefile = gameslot.getGamefile()!;
 	const thisRider = boardutil.getPieceFromCoords(gamefile.pieces, piece.coords)!;
```
Idonotus, line 104: Why do we get thisRider if we are provided piece as a parameter? Is there any case where `thisRider` does not equal `piece`?

Naviary3: Your right, this is probably unnecessary. I'd remove it and then just double check to make sure hovering over arrows still renders their legal moves as expected.

PR #563, opened 2025-05-27
Change onlinegame resignation logic

tsevasa:


translation/en-US.toml
```diff
@@ -593,7 +593,7 @@ ws-too_many_requests_to_server = "Too Many Requests. Try again soon."
 ws-bad_request = "Bad Request"
 ws-not_found = "404 Not Found"
 ws-forbidden = "Forbidden."
-ws-already_in_game = "You are already in a game."
+ws-already_in_game = "You are connected to a game according to the server. This might be caused by large latency, please retry in a few seconds."
```
Naviary2, line 596: This message is intended to appear when you have two tabs open, one of them is in a game, and the other is on the invite creation screen. If you try to click or create an invite on that second screen, this message pops up, but this new message would be confusing.

Perhaps this change may not be needed anymore though if the client requests to be removed from the active games list as soon as they click main menu, as per my comment above.

src/client/scripts/esm/game/gui/guipause.js
```diff
@@ -112,27 +115,44 @@ function updateDrawOfferButton() {
...
 
-	if (!onlinegame.areInOnlineGame() || onlinegame.hasServerConcludedGame()) return element_mainmenu.textContent = translations.main_menu;
+	if (!onlinegame.areInOnlineGame() || onlinegame.hasServerConcludedGame() || onlinegame.hasPlayerPressedAbortOrResignButton() ) {
+		// If the text currently says "Abort Game" or "Resign Game", freeze the button for 1 second in case the user clicked it RIGHT after it switched text! They may have tried to abort or resign and actually not want to exit to main menu.
+		if (freezeMainMenuButtonUponChange && element_mainmenu.textContent !== translations.main_menu) {
+			element_mainmenu.disabled = true;
+			element_mainmenu.classList.add('opacity-0_5');
+			setTimeout(() => {
+				element_mainmenu.disabled = false;
+				element_mainmenu.classList.remove('opacity-0_5');
```
Naviary2, line 145: With this added section there's some newly added redundancy, the code ~10 lines down also temporarily disables the main menu button. We should add a new method for doing that.

src/client/scripts/esm/game/gui/guipause.js
```diff
@@ -184,9 +206,20 @@ function callback_Resume() {
...
 	callback_Resume();
-	gameloader.unloadGame();
 
-	guititle.open();
+	// Unload and exit game immediately if the button text says "Main Menu"
+	if (!is_main_menu_button_used_as_resign_or_abort_button) {
+
+		// Let the onlinegame script know that the player willingly presses the "Main Menu" button.
+		// This can happen if the server has informed him that game has ended or if the player has already pressed the "Resign" or "Abort" during this game.
+		if (onlinegame.areInOnlineGame) onlinegame.onMainMenuButtonPress();
```
Naviary2, line 215: Oop, areInOnlineGame() is a function, not a variable, it will always be truthy like this

src/client/scripts/esm/game/misc/onlinegame/onlinegamerouter.ts
```diff
@@ -142,10 +142,18 @@ function routeMessage(data: WebsocketMessage): void { // { sub, action, value, i
...
 	
-	// These actions are listened to, even when we're not in a game.
-
+	// This action is listened to, even when we're not in a game and even when we don't care about the previous game.
 	if (data.action === 'joingame') return handleJoinGame(data.value);
-	else if (data.action === 'logged-game-info') return handleLoggedGameInfo(data.value);
+
+	// If the player has already willingly pressed the "Main Menu" button after already having pressed the "Resign/Abort" button,
+	// he is no longer interested in seeing anything related to his previous game when the server sends him messages.
+	if (onlinegame.hasPlayerPressedMainMenuButton()) {
```
Naviary2, line 150: This block is unnecessary, as a few lines down we auto exit if we're not in an online game, which in this case is also going to be true if onlinegame.hasPlayerPressedMainMenuButton() is also true.

Naviary2:
When there's latency, when client A resigns, then client B resigns immediately, client B receives the "Another window has connected" message, and it rejoins the game, as the server resends the "joingame" command. Sending the joingame command was originally intended to have client B rejoin the game so they can see that they actually won since client A resigned right before them, as otherwise they would stay on the main menu screen thinking they lost by resignation.

The server no longer needs to send this joingame message though, as client B no longer returns to main menu by pressing resign, so he still sees that he actually won. And, if client B resign and presses main menu *so fast* after client A resigns, then clearly he doesn't care about seeing the results, so in that case I also think we're fine not telling them to "joingame" again.

Where this happens is inside abortresigngame.abortGame() and resignGame(), look for the comments "Resync them to the game because they did not see the game conclusion"

src/server/game/gamemanager/abortresigngame.js
```diff
@@ -50,9 +47,9 @@ function abortGame(ws, game) {
 
 	// Abort
 
-	setGameConclusion(game, 'aborted');
-	onRequestRemovalFromPlayersInActiveGames(ws, game);
 	const opponentColor = typeutil.invertPlayer(colorPlayingAs);
+	setGameConclusion(game, 'aborted');
+	gameutility.sendGameUpdateToColor(game, colorPlayingAs);
 	gameutility.sendGameUpdateToColor(game, opponentColor);
```
Naviary2, line 53: Use sendGameUpdateToBothPlayers() instead

src/server/game/gamemanager/gamemanager.js
```diff
@@ -192,8 +192,12 @@ function onRequestRemovalFromPlayersInActiveGames(ws, game) {
 	// and the game is scheduled to be deleted, just delete it now!
     
 	if (game.deleteTimeoutID === undefined) return; // Not scheduled to be deleted
-	// Is the opponent still in the players in active games list? (has not seen the game results)
+
+	// Is the player still in the players active games list? (has not seen the game results)
 	const color = ws.metadata.subscriptions.game?.color || gameutility.doesSocketBelongToGame_ReturnColor(game, ws);
+	if (!hasColorInGameSeenConclusion(game, color)) return; // They are still in the active games list because they have not seen the game conclusion yet.
```
Naviary2, line 198: I don't believe this has an effect, the method we're in, onRequestRemovalFromPlayersInActiveGames(), is only ever called by the client who has already seen the conclusion (or does not want to). We call removeUserFromActiveGame() a few lines up, that means calling hasColorInGameSeenConclusion() on them is always going to return true.

src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts
```diff
@@ -304,25 +332,40 @@ function reportOpponentsMove(reason: string) {
...
+function onAbortOrResignButtonPress() {
+	if (!inOnlineGame) return;
+	playerHasPressedAbortOrResignButton = true;
+}
+
+/** 
+ * Called when the player presses the "Main Menu" button in an onlinegame
+ * This can happen if the game is already over or if the player has already pressed the "Abort / Resign" button
+ * */
+function onMainMenuButtonPress() {
```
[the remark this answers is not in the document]
tsevasa: You are right, I will do that.

src/client/scripts/esm/game/gui/guipause.js
```diff
@@ -137,36 +137,33 @@ function updateTextOfMainMenuButton({ freezeMainMenuButtonUponChange } = {}) {
...
 		// If the text currently says "Abort Game" or "Resign Game", freeze the button for 1 second in case the user clicked it RIGHT after it switched text! They may have tried to abort or resign and actually not want to exit to main menu.
-		if (freezeMainMenuButtonUponChange && element_mainmenu.textContent !== translations.main_menu) {
-			element_mainmenu.disabled = true;
-			element_mainmenu.classList.add('opacity-0_5');
-			setTimeout(() => {
-				element_mainmenu.disabled = false;
-				element_mainmenu.classList.remove('opacity-0_5');
-			}, 1000);
-		}
+		if (freezeMainMenuButtonUponChange && element_mainmenu.textContent !== translations.main_menu) freezeMainMenuButton(1000);
```
Naviary3, line 140: Thanks for creating freezeMainMenuButton()! Let's also place the freeze duration into a variable.

src/client/scripts/esm/game/gui/guipause.js
```diff
@@ -204,22 +201,19 @@ function callback_Resume() {
 }
 
 function callback_MainMenu() {
-	onlinegame.onMainMenuPress();
+	onlinegame.onAbortOrResignButtonPress();
```
Naviary3, line 204: I think we should only call this if is_main_menu_button_used_as_resign_or_abort_button is true, below

src/server/game/gamemanager/abortresigngame.js
```diff
@@ -31,26 +31,19 @@ function abortGame(ws, game) {
 	if (game.gameConclusion === 'aborted') { // Opponent aborted first.
 		onRequestRemovalFromPlayersInActiveGames(ws, game);
 		return;
-	} else if (gameutility.isGameOver(game)) { // Resync them to the game because they did not see the game conclusion.
-		console.error("Player tried to abort game when the game is already over!");
-		sendNotify(ws, "server.javascript.ws-no_abort_game_over");
-		gameutility.subscribeClientToGame(game, ws, colorPlayingAs);
```
Naviary3, line 40: Same here. We shouldn't let the abort override the existing game conclusion, if the game is already over.

PR #569, opened 2025-06-03
Database modifications and add rating_abuse table

tsevasa:


src/server/database/databaseTables.ts
```diff
@@ -84,9 +85,18 @@ const allGamesColumns: string[] = [
 	'result',
 	'termination',
 	'move_count',
+	'time_duration',
```
Naviary3: Perhaps `time_duration_millis` would be extra clear?

src/server/database/databaseTables.ts
```diff
@@ -176,6 +187,7 @@ function generateTables(): void {
 			game_id INTEGER NOT NULL REFERENCES games(game_id) ON DELETE CASCADE,
 			player_number INTEGER NOT NULL, -- 1 => White  2 => Black
 			score REAL, -- 1 => Win   0.5 => Draw   0 => Loss   NULL => Aborted
+			time_at_end INTEGER, -- Number of milliseconds that player still has left on his clock when the game ended. Null if game has no clock or info is missing.
```
Naviary3: Hmm. Should we do `clock_at_end_millis`? "time" almost feels like it'd be a timestamp

src/server/game/gamemanager/gamemanager.js
```diff
@@ -396,6 +396,9 @@ function onPlayerLostByAbandonment(game, colorWon) {
 async function deleteGame(game) {
 	if (!game) return console.error(`Unable to delete an undefined game!`);
 
+	// The ending time of the game object is set now as the game is deleted.
+	game.timeEnded = Date.now();
```
Naviary3: I think this would be better placed inside setGameConclusion or onGameConclusion(), that way the timeEnded is the same time at which we stop the clocks.

Although maybe if a cheat report comes in after set/onGameConclusion has been called once before already, then we don't need to init timeEnded again..

src/server/game/TypeDefinitions.js
```diff
@@ -66,6 +66,8 @@ function Game() {
 	this.id = undefined;
 	/** The time this game was created. The number of milliseconds that have elapsed since the Unix epoch. */
 	this.timeCreated = undefined;
+	/** The time this game ended. The number of milliseconds that have elapsed since the Unix epoch. */
+	this.timeEnded = undefined;
```
Naviary3, line 70: After that we could explicitly specify that this is the time the clocks are stopped.

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -106,6 +106,7 @@ async function enterGameInGamesTable(game: Game, dateSqliteString: string, ratin
 	const leaderboard_id = VariantLeaderboards[game.variant] ?? null; // Include the leaderboard_id even if the game wasn't rated, so we can still filter
 	const game_private: 0 | 1 = (game.publicity !== 'public' ? 1 : 0);
 	const { base_time_seconds, increment_seconds } = clockutil.splitTimeControl(game.clock);
+	const game_time_duration_millis = (game.timeCreated !== undefined && game.timeEnded !== undefined ? game.timeEnded - game.timeCreated : null);
```
Naviary3: I think timeCreated is always gonna be defined, even for migrating game logs

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -228,6 +230,10 @@ async function updatePlayerGamesTable(game: Game, game_id: number, victor: Playe
 		if (user_id === undefined) continue; // Guest players don't get an entry in the player_games table or an elo for updating
 
 		const score = victor === undefined ? null : victor === player ? 1 : victor === players.NEUTRAL ? 0.5 : 0;
+
+		const ending_clocks = gameutility.getGameClockValues(game)?.clocks;
```
Naviary3: Move this out of the for loop so we don't fetch it once for every player

src/server/game/gamemanager/gamelogger.ts
```diff
@@ -228,6 +230,10 @@ async function updatePlayerGamesTable(game: Game, game_id: number, victor: Playe
...
+		const time_at_end = (ending_clocks !== undefined && ending_clocks[playerStr] !== undefined ? ending_clocks[playerStr] : null);
```
Naviary3: ending_clocks will always be defined, just not the players within it if the game is untimed

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -68,11 +69,26 @@ async function measureRatingAbuseAfterGame(game: Game) {
...
 
+	// Access the player rating_abuse data
+	const rating_abuse_data = getRatingAbuseData(user_id, leaderboard_id, ['game_count_since_last_check', 'last_alerted_at']);
+	if (rating_abuse_data === undefined) {
+		await logEventsAndPrint(`Unable to read rating_abuse_data of user ${user_id} on leaderboard ${leaderboard_id} while making RatingAbuse check!`, 'errLog.txt');
+		return;
+	}
+
+	// Increment game_count_since_last_check by 1
+	const game_count_since_last_check = 1 + (rating_abuse_data.game_count_since_last_check ? rating_abuse_data.game_count_since_last_check : 0);
```
Naviary3: This is a little more concise:
```js
1 + (rating_abuse_data.game_count_since_last_check || 0)
```

Naviary3: Ah also, I was thinking `game_count_since_last_check` would be reset to 0 whenever we did a check

src/server/database/ratingAbuseManager.ts
```diff
@@ -0,0 +1,219 @@
...
+		// Log the error for debugging purposes
+		logEventsAndPrint(`Error adding entry to rating_abuse table for user "${user_id}" and leaderboard "${leaderboard_id}": ${message}`, 'errLog.txt');
+
+		// Return an error message
+		// Check for specific constraint errors if possible (e.g., FOREIGN KEY failure)
+		let reason = 'Failed to add entry to rating_abuse table.';
+		if (error instanceof Error && 'code' in error) {
+			// Example check for better-sqlite3 specific error codes
+			if (error.code === 'SQLITE_CONSTRAINT_FOREIGNKEY') {
+				reason = '(User ID, Leaderboard ID) does not exist in the rating_abuse table.';
```
Naviary3: Do you mean does not exist in the leaderboards table?

src/server/database/ratingAbuseManager.ts
```diff
@@ -0,0 +1,219 @@
...
+			} else if (error.code === 'SQLITE_CONSTRAINT_UNIQUE' || error.code === 'SQLITE_CONSTRAINT_PRIMARYKEY') {
+				reason = '(User ID, Leaderboard ID) already exists in the rating_abuse table.';
+			}
+		}
+		return { success: false, reason };
+	}
+}
+
+/**
+ * Checks if an entry exists in the rating_abuse.
```
Naviary3: rating_abuse table?

src/server/database/ratingAbuseManager.ts
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+
+	// Dynamically build the SET part of the query
+	const setStatements = Object.keys(columnsAndValues).map(column => `${column} = ?`).join(', ');
+	const values = Object.values(columnsAndValues);
+
+	// Add the user_id and leaderboard_id as the last parameters for the WHERE clause
+	values.push(user_id);
+	values.push(leaderboard_id);
```
Naviary3: These can both be pushed with one call

src/server/database/ratingAbuseManager.ts
```diff
@@ -0,0 +1,219 @@
...
+		// Check if the update was successful
+		if (result.changes > 0) return { success: true, result };
+		else {
+			logEventsAndPrint(`No changes made when updating rating_abuse table columns ${JSON.stringify(columnsAndValues)} for entry in rating_abuse table with user ID "${user_id}" and leaderboard ID "${leaderboard_id}"! Received: ${jsutil.ensureJSONString(columnsAndValues)}`, 'errLog.txt');
+			return { success: false, reason: 'No changes made.' }; // Generic error message
+		}
+	} catch (error: unknown) {
+		const message = error instanceof Error ? error.message : String(error);
+		// Log the error for debugging purposes
+		logEventsAndPrint(`Error updating rating_abuse table columns ${JSON.stringify(columnsAndValues)} for user ID "${user_id}" and leaderboard ID "${leaderboard_id}": ${message}! Received: ${jsutil.ensureJSONString(columnsAndValues)}`, 'errLog.txt');
```
Naviary3: This stringifies columnsAndValues into the error message twice.. is the first one supposed to be just the keys stringified?

src/server/database/databaseTables.ts
tsevasa: Good idea!

src/server/database/databaseTables.ts
tsevasa: Good idea, changing this too.

src/server/game/gamemanager/gamemanager.js
tsevasa: Hm, good idea, I will move this to onGameConclusion()

tsevasa: and make sure that it is only set the first time

src/server/game/gamemanager/ratingabuse.ts
tsevasa: Ah, I was thinking we just keep incrementing it forever. Cannot hurt and it's just as easy to check if it's zero mod some number. But whatever you prefer.

tsevasa: You are right, I will change it. Else the name doesn't make any sense.

PR #570, opened 2025-06-04
Correctly show username containers in local games

tsevasa:


src/server/controllers/createAccountController.ts
```diff
@@ -39,6 +39,8 @@ import { logEventsAndPrint } from '../middleware/logEvents.js';
  * admin
  */
 const reservedUsernames: string[] = [
+	'White', 'Black', 'Engine', 'Guest', 'You',
+	'white', 'black', 'engine', 'guest', 'you',
```
Naviary2, line 43: These aren't case sensitive so lowercase is fine.

Also, "Guest" and "You" are already taken...

PR #571, opened 2025-06-04
Add some rating abuse checks

tsevasa:


src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,103 @@
...
+const NUMBER_OF_SUSPICIOUS_ENTRIES_TO_RAISE_ALARM = 3;
+
+/** Buffer time for sending the next email. If a user is found suspicious several times in that interval, no email is sent. */
+const SUSPICIOUS_USER_NOTIFICATION_BUFFER_MILLIS = 1000 * 60 * 60 * 24; // 24 hours
+
+/** Number of rated game pairs started close after each other to count as suspicious. */
+const TOO_CLOSE_GAMES_AMOUNT = 2;
+
+/** Two rated games started this close after each other count as suspicious. */
+const TOO_CLOSE_GAMES_MILLIS = 1000 * 60 * 10; // 10 minutes
```
Naviary2: If the user plays exclusively bullet games then they can get several in within 10 minutes legitimately. Maybe we should instead rely on move count and clock_at_end to second-handedly measure too close together games?

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,103 @@
...
+
+/** Games with fewer moves than this are suspicious. */
+const SUSPICIOUS_MOVE_COUNT = 10;
```
Naviary2: How about we use a graph to weigh the total suspicion level.
https://www.desmos.com/calculator/6fwjozlvxq

The average move count per game is 38. In this graph, a game of 25 (some cushion) moves would add 0 weight to their total suspicion level. While a game of 2 moves would add 1 point of suspicion. We could further divide the sucpicion per game by the total number of games analyzed. So say we analyzed their 10 past games, they would only receive 1 point of suspicion if *all 10* of those games were 2 moves long. But say if 9 of those games were 25+ moves, and only **1** game was 2 moves, they would only receive 0.1 points of suspicion. You can't control if your opponent decides to resign early, after all, it only becomes more suspicious if it's consistent.

Then we could total up the supicion level of each category, and if the number exceeds 1 we send an alert email. How's that?

Naviary2: I added math.getYAtX_GeneralForm(). Now we can defined two customizable points (2,1) and (25,0), use math.getLineGeneralFormFrom2Coords() to get the line equation, then use getYAtX_GeneralForm passing in the number of moves to get the suspicion level of any one game.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,103 @@
...
+
+/** Games lasting less than this time on the serverare suspicious. */
+const SUSPICIOUS_TIME_DURATION_MILLIS = 1000 * 60; // 1 minute
```
Naviary2: For same reason above, because of bullet games, the time spent per move, and thus game duration, can vary widely while still being legitimate. Perhaps we no longer measure this and just rely on move count and clock remaining at end to determine if they didn't spend enough time thinking.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,103 @@
...
+
+/** A player ending a game with a larger fraction of his total clock time than this counts as suspicious. */
+const SUSPICIOUS_CLOCK_REMAINING_FRACTION = 0.9;
```
Naviary2: For remaining clock time we could maybe use this graph to weigh their suspicion
https://www.desmos.com/calculator/tm1nngxovp

Where the x axis is their percentage time remaining at end of their total time.

(Clamp to 0-1)

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -91,25 +158,130 @@ async function measurePlayerRatingAbuse(user_id: number, leaderboard_id: number)
...
+		['game_id', 'date', 'base_time_seconds', 'increment_seconds', 'private', 'termination', 'move_count', 'time_duration_millis']
+	) as RatingAbuseRelevantGamesRecord[];
+	const games_table_game_id_list = recentGamesEntries.map(recent_game => recent_game.game_id);
+
+	// Combine the information about the games into a single gameInfoList object
+	const gameInfoList: RatingAbuseRelevantGameInfo[] = [];
+	for (let i = 0; i < game_id_list.length; i++) {
+		const j = games_table_game_id_list.indexOf(game_id_list[i]!);
+		// If the same game_id exists in both lists of retrieved database entries, add this game as a single object to gameInfoList
+		if (j > -1) gameInfoList.push({ ...recentPlayerGamesEntries[i]!, ...recentGamesEntries[j]! });
```
Naviary2, line 185: This is a cause for error if the game id is found in player_games but not games, we should log it.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,103 @@
...
+};
+
+/** Object containing all relevant information about a specific game, which is used for the rating abuse calculation */
+type RatingAbuseRelevantGameInfo = RatingAbuseRelevantPlayerGamesRecord & RatingAbuseRelevantGamesRecord;
+
+/** Object containing information about analysis of suspicion level of some characteristic */
+type SuspicionLevelRecord = {
+	game_id?: number,
+	suspicion_level: number,
+	reason?: string
```
Naviary2: What if instead of analyzing one game at a time's suspicion level, we analyzed their most recent N games for one category at a time.

```js
type SuspicionLevelRecord = {
	category: 'move_count' | 'clock_at_end',
	weight: number,
}
```

For example, for the "move_count" category, we sum up the total weight of the last N games, using the graph above, then take the average, and add it to suspicion_level_record_list as a SuspicionLevelRecord

This way, when the alert emails are sent, I immediately can see what categories raised the most suspicion. For example if move_count had a weight of 0.9 and clock_at_end had a weight of 0.3, I would first review the move count of each game to verify if the player was cheating or not.

It also means the number of SuspicionLevelRecords don't increase linearly with game count analyzed.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: How about keeping this and reducing the time threshhold to something more extreme? Something so short might warrant a slight suspicion increase in any case.

I also implemented a linear interpolation for this case.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: How about keeping this and reducing the time threshhold to something more extreme? Something so short might warrant a slight suspicion increase in any case.

I also implemented a linear interpolation for this case.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: Sounds good! I've implemented this now.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: Sounds good, I've implemented this now.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: I really like the idea of the linear interpolation, but math.getYAtX_GeneralForm() seems like complete overkill to me.

I now implemented this interpolation far more directly (and simpler, in my opinion) directly in ratingabuse.ts.

src/server/game/gamemanager/ratingabuse.ts
Naviary3: Great! In the future we can actually calculate the average time remaining at the end of games to get this more accurate.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -91,25 +151,164 @@ async function measurePlayerRatingAbuse(user_id: number, leaderboard_id: number)
...
+	else logEvents(`INNOCENT! Ran suspicion check for user (${user_id}) on leaderboard (${leaderboard_id}) with net rating change ${netRatingChange} in the last ${GAME_INTERVAL_TO_MEASURE} games, but user is not suspicious. Suspicion total weight: ${suspicion_total_weight}. Game IDs: ${JSON.stringify(game_id_list)}. Suspicion list: ${JSON.stringify(suspicion_level_record_list)}`, 'ratingAbuseLog.txt');
+}
+
+
+/**
+ * Check if the game dates are too close in proximity to each other
+ * If yes, append entry to suspicion_level_record_list.
+ */
+function checkCloseGamePairs(gameInfoList: RatingAbuseRelevantGameInfo[], suspicion_level_record_list: SuspicionLevelRecord[]) {
+	const sorted_timestamp_list = gameInfoList.map(game_info => game_info.date).map(date => timeutil.sqliteToTimestamp(date)).sort();
```
Naviary3: We can merge the map calls into one.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -91,25 +151,164 @@ async function measurePlayerRatingAbuse(user_id: number, leaderboard_id: number)
...
+	const timestamp_differences: number[] = [];
+	for (let i = 1; i < sorted_timestamp_list.length; i++) {
+		timestamp_differences.push(sorted_timestamp_list[i]! - sorted_timestamp_list[i - 1]!);
+	}
+	const close_game_pairs_amount = timestamp_differences.filter(diff => diff < TOO_CLOSE_GAMES_MILLIS).length;
```
Naviary3, line 236: The fastest time control, 1+1, at 38 average moves per game, means a legitimate game could be as short as 2m16s (probably shorter since there's always time left over on atleast one person's clock. Currently, if their last 5 games came 4m30s apart they are flagged as 100% suspicious. I think this needs to be much lower.

What if the buffer was 2 minutes?

Naviary3, line 235: The fastest time control, 1+1, at 38 average moves per game, means a legitimate game could be as short as 3m16s (probably shorter since there's always time left over on atleast one person's clock. Currently, if their last 5 games came 4m30s apart they are flagged as 100% suspicious. I think this needs to be lower.

Considering it also takes some seconds to join a new game.. what if the buffer was instead 4 minutes?

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -11,36 +11,96 @@
...
+const SUSPICIOUS_USER_NOTIFICATION_BUFFER_MILLIS = 1000 * 60 * 60 * 24; // 24 hours
+
+/** Two rated games started this close after each other have a nonzero suspicion score. */
+const TOO_CLOSE_GAMES_MILLIS = 1000 * 60 * 5; // 5 minutes
+
+/** Games with fewer moves than this have a nonzero suspicion score. */
+const SUSPICIOUS_MOVE_COUNT = 25;
+
+/** Games lasting less than this time on the server have a nonzero suspicion score. */
+const SUSPICIOUS_TIME_DURATION_MILLIS = 1000 * 60; // 60 seconds
```
Naviary3: As the above comment, the avg time for a 1+1 game could be 3m16s. Perhaps we set SUSPICIOUS_TIME_DURATION_MILLIS at 3 minutes? Or 2.5 minutes?

tsevasa: Sure, let's go with 3 minutes for now!

src/server/game/gamemanager/ratingabuse.ts
tsevasa: Fair enough, let's reduce the buffer!

PR #574, opened 2025-06-08
Update Move format to be more compatable with server side

Idonotus:


src/client/scripts/esm/chess/util/moveutil.ts
```diff
@@ -248,6 +248,11 @@ function stripSpecialMoveTagsFromCoords(coords: CoordsSpecial): Coords {
 }
 
 
+function isMoveSelf(move: _Move_Compact): boolean {
```
Naviary3, line 251: Add jsdoc indicating this method tests for null moves.

src/client/scripts/esm/chess/util/moveutil.ts
```diff
@@ -248,6 +248,11 @@ function stripSpecialMoveTagsFromCoords(coords: CoordsSpecial): Coords {
...
+	return move.startCoords[0] === move.endCoords[0] && move.startCoords[1] === move.endCoords[1];
+}
```
Naviary3, line 257: Add jsdoc indicating this method tests for null moves.

src/client/scripts/esm/chess/logic/movepiece.ts
```diff
@@ -316,9 +265,15 @@ function queueIncrementMoveRuleStateChange({ basegame, boardsim }: FullGame, mov
...
  * Executes all the logical board changes of a global forward move in the game, no graphical changes.
  */
-function makeMove(gamefile: FullGame, move: Move | NullMove) {
+function makeMove(gamefile: FullGame, move: Move) {
 	gamefile.boardsim.moves.push(move);
-	gamefile.basegame.moves.push(move.isNull ? '--' : move.compact);
+	gamefile.basegame.moves.push({
+		startCoords: move.startCoords,
+		endCoords: move.endCoords,
+		promotion: move.promotion,
```
Naviary3, line 273: Can be simplified with `{ ...move }`

src/client/scripts/esm/chess/util/moveutil.ts
```diff
@@ -71,14 +71,14 @@ function isDecrementingLegal(boardsim: Board): boolean {
 /**
  * Tests if the provided index is out of range of the moves list length
  */
-function isIndexOutOfRange(moves: (Move | NullMove)[], index: number): boolean {
+function isIndexOutOfRange(moves: any[], index: number): boolean {
```
Naviary3: Let's avoid using "any":
```ts
moves: Move[] | BaseMove[]
```
is there any other array type we'd pass into here?

Naviary3: For the heck of it, maybe even this would work since Move contains all of BaseMove's properties already:
```ts
moves: BaseMove[]
```
if typescript gets angry then we can go with the above option.

src/client/scripts/esm/chess/util/moveutil.ts
```diff
@@ -112,14 +112,14 @@ function areWeViewingLatestMove(boardsim: Board): boolean {
...
-function isIndexTheLastMove(moves: (Move | NullMove)[], index: number): boolean {
+function isIndexTheLastMove(moves: Move[], index: number): boolean {
 	const finalIndex = moves.length - 1;
 	return index === finalIndex;
 }
 
 /**
  * Gets the color of whos turn it is currently, or at the front of the game.
- * Depends on the turn order.
+ * Depends on the turn order. WILL NOT ACCOUNT FOR NULL MOVES.
```
Naviary3, line 122: Does it not, though? Null moves still flip the turn order. basegame.moves is exactly the same length as boardsim.moves, still pushing null moves to it, right?

src/client/scripts/esm/game/misc/onlinegame/movesendreceive.ts
```diff
@@ -37,7 +37,7 @@ function sendMove() {
 
 	const gamefile = gameslot.getGamefile()!;
 	const lastMove = moveutil.getLastMove(gamefile.boardsim.moves)!;
-	if (lastMove.isNull) throw Error('Cannot submit null move to online game.');
+	if (moveutil.isMoveSelf(lastMove)) throw Error('Cannot submit null move to online game.');
```
Naviary3: We could prob just delete this. Users can't realistically make null moves, and if they do then legal move validation will catch it, whether on the server in the future, or the other client sees it and reports for cheating.

src/client/scripts/esm/game/misc/onlinegame/resyncer.ts
```diff
@@ -91,7 +91,7 @@ function synchronizeMovesList(gamefile: FullGame, mesh: Mesh | undefined, moves:
 	const finalMoveIsOurMove = boardsim.moves.length > 0 && moveutil.getColorThatPlayedMoveIndex(gamefile.basegame, boardsim.moves.length - 1) === onlinegame.getOurColor();
 	const previousMove = boardsim.moves.length > 1 ? boardsim.moves[boardsim.moves.length - 2] : undefined;
 	const previousMoveMatches = (moves.length === 0 && boardsim.moves.length === 1)
-		|| (boardsim.moves.length > 1 && moves.length > 0 && !previousMove!.isNull && previousMove!.compact === moves[moves.length - 1]!.compact);
+		|| (boardsim.moves.length > 1 && moves.length > 0 && !moveutil.isMoveSelf(previousMove!) && previousMove!.compact === moves[moves.length - 1]!.compact);
```
Naviary3: We can delete the `!moveutil.isMoveSelf(previousMove!)` check now. I think we only needed it so typescript was confident some property of Move was present.

src/client/scripts/esm/game/misc/onlinegame/resyncer.ts
```diff
@@ -115,7 +115,7 @@ function synchronizeMovesList(gamefile: FullGame, mesh: Mesh | undefined, moves:
 	while (true) { // Decrement i until we find the latest move at which we're in sync, agreeing with the server about.
 		if (i === -1) break; // Beginning of game
 		const thisGamefileMove = boardsim.moves[i];
-		if (thisGamefileMove && !thisGamefileMove.isNull) { // The move is defined
+		if (thisGamefileMove && !moveutil.isMoveSelf(thisGamefileMove)) { // The move is defined
```
Naviary3: Same here

src/client/scripts/esm/chess/logic/movepiece.ts
Naviary3, line 1: We can delete the commented-out ensureMovesNotNull() method now, and delete the commented out export at the bottom of the file.

src/client/scripts/esm/chess/util/moveutil.ts
Idonotus: but this is a generic method, it won't really matter what we put in as long as its an array. I would think this could be used elswhere

Naviary2: If it's generic to other arrays, then -1 is an issue because that index is considered valid.
It's also a moveutil script, it's just meant to work with moves list in the gamefile.
Until we are able to use it elsewhere I still think we should make it more type safe.

src/client/scripts/esm/chess/logic/movepiece.ts
```diff
@@ -316,9 +265,12 @@ function queueIncrementMoveRuleStateChange({ basegame, boardsim }: FullGame, mov
 /**
  * Executes all the logical board changes of a global forward move in the game, no graphical changes.
  */
-function makeMove(gamefile: FullGame, move: Move | NullMove) {
+function makeMove(gamefile: FullGame, move: Move) {
 	gamefile.boardsim.moves.push(move);
-	gamefile.basegame.moves.push(move.isNull ? '--' : move.compact);
+	gamefile.basegame.moves.push({
+		...move
+	});
```
Naviary3, line 275: Note to self: Revert this to what it was before. I missed that would spread all properties of a Move type, so that would put unused properties into the basegame move, that's my bad.

src/client/scripts/esm/chess/util/moveutil.ts
```diff
@@ -247,6 +247,14 @@ function stripSpecialMoveTagsFromCoords(coords: CoordsSpecial): Coords {
 	return coordutil.copyCoords(coords); // Does not copy non-enumerable properties
 }
 
+/**
+ * Tests if a move is null.
```
Naviary3: Note to self: Also comment that engines should be the only entities to create and make null moves.

Naviary3:
Looks good! I'll merge this when I'm home later and to a final compilation check.

PR #575, opened 2025-06-10
Prevent account deletion during online game

tsevasa:


src/server/game/gamemanager/activeplayers.js
```diff
@@ -107,7 +107,16 @@ function hasColorInGameSeenConclusion(game, color) {
...
+	if (player.member)  return membersInActiveGames[player.member];
+}
+
+/**
+ * Checks if member with a given username is currently listed as being in some active game
+ * @param {string} username - username of some member
+ * @returns {boolean} true if member is currently in active game, otherwise false
+ */
+function isMemberInSomeActiveGame(username) {
+	return username in membersInActiveGames;
```
Naviary3: So a user may not be in the membersInActiveGames list, but still be in a game. membersInActiveGames more-so controls who is allowed to join a new game, as a user is removed from this list as soon as they've seen the conclusion from the game. The game itself may be deleted seconds later, so we still may get a crash if they delete their account immediately after seeing the game results, but *before* the opponent sees the result.

We can test by making the opponent (player B) leave the page right before player A ends the game. Then player A deletes their account before the 8-second timer is up, and the game is deleted. I predict there will be a crash in that scenario.

Naviary3: We probably have to have gamemanager iterate through all activeGames to see if the player is ACTUALLY in a live game.

src/server/controllers/deleteAccountController.js
```diff
@@ -18,33 +19,41 @@ import { closeAllSocketsOfMember } from '../socket/socketManager.js';
...
+	const { user_id, username } = getMemberDataByCriteria(['user_id', 'username'], 'username', claimedUsername);
+	if (user_id === undefined || username === undefined) {
 		return logEventsAndPrint(`Unable to find member of claimed username "${claimedUsername}" after a correct password to delete their account!`, 'errLog.txt');
 		// if (user_id === undefined) return logEventsAndPrint(`User "${usernameCaseInsensitive}" not found after a successful login! This should never happen.`, 'errLog.txt');
 	}
+
+	// Do not allow account deletion if user is currently playing a game
+	if (isMemberInSomeActiveGame(username)) {
+		logEventsAndPrint(`User ${username} requested accont deletion while being listed in some active game.`, 'deletedAccounts.txt');
+		return res.status(404).json({ 'message' : getTranslationForReq("server.javascript.ws-deleting_account_in_game", req) });
```
Naviary3: Maybe code 403 (Forbidden) fits better?

src/server/game/gamemanager/gamemanager.js
```diff
@@ -107,6 +107,20 @@ function addGameToActiveGames(game) {
...
 
+/**
+ * Checks if member with a given username is currently listed as being in some active game
+ * @param {string} username - username of some member
+ * @returns {boolean} true if member is currently in active game, otherwise false
+ */
+function isMemberInSomeActiveGame(username) {
+	for (const game of Object.values(activeGames)) {
+		for (const player of Object.values(game.players)) {
+			if (player.identifier?.member === username) return true;
```
Naviary3: Minor: identifier should always be defined, but not member. Should be fine though as long as we never pass in an undefined username.

Everything looks good!

tsevasa: Ok, I replaced the `?.` with a `.`

Yes, `username` is never supposed to be undefined.

PR #577, opened 2025-06-11
Add some more rating abuse checks (IP address and repeat opponent checking)

tsevasa:


src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -187,6 +190,34 @@ async function measurePlayerRatingAbuse(user_id: number, leaderboard_id: number)
...
+	const unique_user_id_list = [...new Set(user_id_list)];
+
+	// Dictionary of frequencies of user_ids in user_id_list
+	const user_id_frequency: { [key: number] : number } = {};
+	for (const user_id of user_id_list) {
+		user_id_frequency[user_id] = (user_id_frequency[user_id] || 0) + 1;
+	}
+
+	// Get the refresh tokens of the user and all his opponents
+	const refreshTokenEntries = findRefreshTokensForUsers([user_id].concat(unique_user_id_list));
```
Naviary2: This needs to be surrounded with try-catch since refreshTokenManager isn't in charge of catching database errors.

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -195,18 +226,26 @@ async function measurePlayerRatingAbuse(user_id: number, leaderboard_id: number)
...
 	
 	// Run various checks and add entries to suspicion_level_record_list, if necessary
-	checkCloseGamePairs(gameInfoList, suspicion_level_record_list);
-	checkMoveCounts(gameInfoList, suspicion_level_record_list);
-	checkDurations(gameInfoList, suspicion_level_record_list);
-	checkClockAtEnd(gameInfoList, suspicion_level_record_list);
+	try {
+		checkCloseGamePairs(gameInfoList, suspicion_level_record_list);
+		checkMoveCounts(gameInfoList, suspicion_level_record_list);
+		checkDurations(gameInfoList, suspicion_level_record_list);
```
Naviary2: Was there a specific crash you're worried about with this try-catch? At a quick glance I don't see anything that would throw an error in these..

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -310,6 +349,57 @@ function checkClockAtEnd(gameInfoList: RatingAbuseRelevantGameInfo[], suspicion_
...
+
+	let weight = 0;
+	for (const user_id in opponent_ip_address_lists) {
+		// Player is suspicious if he uses a same IP adress as an opponent
+		const common_ip_addresses = user_ip_address_list.filter(ip_address => opponent_ip_address_lists[user_id]!.includes(ip_address));
+		if (common_ip_addresses.length > 0) weight += user_id_frequency[user_id] ?? 0;
+	}
+	if (weight > 0) suspicion_level_record_list.push({
+		category: 'ip_addresses',
+		weight: 2 * weight / user_id_list.length // rescale to [0,2]
```
Naviary2: Also here- What if someone is playing with a sibling in the same household? Same ip, but legitimate games. For that reason I don't think all recent games with matching ip addresses should yield 1 (or is it 2 here?) point, maybe 0.5 or 0.75?

src/server/game/gamemanager/ratingabuse.ts
```diff
@@ -310,6 +349,57 @@ function checkClockAtEnd(gameInfoList: RatingAbuseRelevantGameInfo[], suspicion_
...
+	if (user_id_list.length === 0) return;
+
+	let weight = 0;
+	for (const frequency of Object.values(user_id_frequency)) {
+		// Player is suspicious if he played against the same opponent several times
+		if (frequency > 1) weight += frequency ** 2;
+	}
+	if (weight > 0) suspicion_level_record_list.push({
+		category: 'same_opponents',
+		weight: weight / (user_id_list.length ** 2) // rescale to [0,1]
```
Naviary2, line 394: I don't think someone's past 5 games being played with the same person should give 1 point. Ideally only if there's also other flags present does the total weight exceed 1. Maybe it should give 0.5, or 0.75 points, if all their recent games are with the same person?

src/server/game/gamemanager/ratingabuse.ts
tsevasa: ok, done!

src/server/game/gamemanager/ratingabuse.ts
tsevasa: To me it felt like one of the strongest indicators, since playing against a sibling from the same household would be much rarer than just cheating, that's why I upscaled it to 2 - for the extra scrutiny. How about I just put it back down to 1 again? It won't be enough to trigger a cheat report by itself, after all.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: No, I never got a crash here, but I was just a bit paranoid since I wanna avoid server crashes out of unforeseen reasons.

I removed this try-catch and just put a guard clause around the entire `measurePlayerRatingAbuse()` function call - cannot hurt!

src/server/game/gamemanager/ratingabuse.ts
tsevasa: Ok, I reduced it to 1 again.

src/server/game/gamemanager/ratingabuse.ts
tsevasa: I put this equal to 1 here, because I raised `SUSPICION_TOTAL_WEIGHT_THRESHHOLD` to 2 in this PR.

My thinking was that `SUSPICION_TOTAL_WEIGHT_THRESHHOLD` is too low, if it's only equal to 1, since I fear it would get triggered far too often, if even a single of the other checks has a high weight (especially if we keep adding more checks). That's why I raised it to 2 in this PR, so that no single check can raise alarms but only a combination of them can do it.

PR #589, opened 2025-06-18
ThisIsNotAvailable's Premoves

Naviary2:


src/client/scripts/esm/game/chess/premoves.ts
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return boardCopy;
+}
+
+function showPremoves(premoves : MoveDraft[]) {
+	const gamefile = gameslot.getGamefile();
+	if (!gamefile || !premoves || premoves.length === 0) return;
+
+	// Always start from a fresh deep copy
+	const boardCopy : FullGame = jsutil.deepCopyObject(gamefile);
```
Naviary2, line 100: Same thing here

src/client/scripts/esm/game/chess/selection.ts
```diff
@@ -270,9 +288,28 @@ function testIfPieceMoved(gamefile: FullGame, mesh: Mesh | undefined): void {
...
+			guipromotion.open(color);
+			perspective.unlockMouse();
+			pawnIsPromotingOn = hoverSquare;
+			return;
+		}
+
+		premoves.rewindPremovesVisuals();
+		premoves.addPremove(moveDraft, pieceSelected);
+		unselectPiece();
+		premoves.applyPremovesVisuals();
```
Naviary3: premoves.ts should be the one to automatically rewindPremovesVisuals() before adding a new premove and applyPremovesVisuals() afterward.

Actually that's probably not even needed, when we addPremove() all we have to do it apply the board changes of that one premove, its changes should come last anyway.

src/client/scripts/esm/game/chess/selection.ts
```diff
@@ -322,8 +359,9 @@ function canMovePieceType(pieceType: number): boolean {
 	if (editMode) return true; // Edit mode allows pieces to be moved on any turn.
 	const isOpponentPiece = isOpponentType(gameslot.getGamefile()!.basegame, pieceType);
 	if (isOpponentPiece) return false; // Don't move opponent pieces
-	const isPremove = !isOpponentPiece && !gameloader.areInLocalGame() && !gameloader.isItOurTurn();
-	return (!isPremove); // For now we can't premove, can only move our pieces on our turn.
+	// const isPremove = !isOpponentPiece && !gameloader.areInLocalGame() && !gameloader.isItOurTurn();
+	// return (!isPremove); // For now we can't premove, can only move our pieces on our turn.
+	return true;
```
Naviary3, line 385: We should only be able to premove if preferences.getPremoveMode() is true.
We'll have to enable that setting toggle now.

PR #597, opened 2025-07-02
Board editor: Attempt number 2

tsevasa:


src/client/scripts/esm/chess/logic/organizedpieces.ts
```diff
@@ -453,8 +453,11 @@ function calcRemainingExistingTypes(positionExistingTypes: Set<number>, turnOrde
 	let existingTypes: number[];
 	let existingRawTypes: RawType[];
 	if (editor) {
-		// ALL pieces may be added in the board editor
-		existingTypes = typeutil.buildAllTypesForPlayers(Object.values(players), Object.values(rawTypes));
+		// ALL pieces may be added in the board editor, but only of the players mentioned in turnOrder
+		const playersSet: Set<Player> = new Set(turnOrder);
+		const playersArray: Array<Player> = [...playersSet];
+		existingTypes = typeutil.buildAllTypesForPlayers(playersArray, Object.values(rawTypes));
+		existingTypes = [...neutralRawTypes, ...existingTypes];
```
Naviary3: Is adding all of the neutral types intended for gargoyles in multiplayer games? If so, we should only need them when player 3 or higher is present (look at the logic below)

src/client/css/play.css
```diff
@@ -185,6 +185,8 @@ canvas {
     position: absolute;
     width: 100%;
     height: 100%;
+    display: flex;
+    flex-direction: column;
```
Naviary3, line 190: Good to comment above these that they are required for the board editor UI

src/client/css/play.css
```diff
@@ -294,6 +296,10 @@ canvas {
     background: linear-gradient(to bottom, white, rgb(255, 255, 255), white);
 }
 
+.active {
```
Naviary3, line 299: Let's also add a section in this stylsheet for storing the board editor specific classes, even though it'll just have this one item for now.

src/client/css/play.css
```diff
@@ -294,6 +296,10 @@ canvas {
...
+    background-color: green;
+}
```
Naviary3: Can this be moved below /* Board Editor UI */ ?

src/client/scripts/esm/chess/logic/state.ts
```diff
@@ -140,15 +141,15 @@ interface EnPassant {
 
 
 /** Creates a check local StateChange, adding it to the Move and immediately applying it to the gamefile. */
-function createCheckState(move: Move, current: inCheck, future: inCheck, gamestate: GameState) {
+function createCheckState(move: Move | Edit, current: inCheck, future: inCheck, gamestate: GameState) {
```
Naviary3: We now have some redundancy between the `Edit` and `Move` types. I may opt for moving `Edit` to movepiece.ts and modifying `Move` to extend `Edit`.

Then boardeditor.ts can import `Edit`,
and this line can just accept an Edit instead of Move | Edit, typescript shouldn't get upset since Move would then contain all properties Edit requires.

src/client/scripts/esm/chess/logic/state.ts
```diff
@@ -140,15 +141,15 @@ interface EnPassant {
...
 	const newStateChange: StateChange = { type: 'check', current, future };
 	move.state.local.push(newStateChange); // Check is a local state
 	// Check states are immediately applied to the gamefile
 	applyLocalState(gamestate.local, newStateChange, true);
 }
 
 /** Creates an attackers local StateChange, adding it to the Move and immediately applying it to the gamefile. */
-function createAttackersState(move: Move, current: Attacker[], future: Attacker[], gamestate: GameState) {
+function createAttackersState(move: Move | Edit, current: Attacker[], future: Attacker[], gamestate: GameState) {
```
Naviary3: same here

src/client/scripts/esm/chess/logic/state.ts
```diff
@@ -171,14 +172,14 @@ function createEnPassantState(move: Move, current?: EnPassant, future?: EnPassan
 }
 
 /** Creates a specialrights global StateChange, queueing it by adding it to the Move. */
-function createSpecialRightsState(move: Move, coordsKey: CoordsKey, current: boolean, future: boolean) {
+function createSpecialRightsState(move: Move | Edit, coordsKey: CoordsKey, current: boolean, future: boolean) {
```
Naviary3: same here

src/client/scripts/esm/chess/logic/state.ts
```diff
@@ -171,14 +172,14 @@ function createEnPassantState(move: Move, current?: EnPassant, future?: EnPassan
...
 	if (current === future) return; // If the current and future values are identical, we can skip queueing this state.
 	const newStateChange: StateChange = { type: 'specialrights', current, future, coordsKey };
 	move.state.global.push(newStateChange); // Special Rights is a global state
 }
 
 /** Creates a moverule global StateChange, queueing it by adding it to the Move. */
-function createMoveRuleState(move: Move, current: number, future: number) {
+function createMoveRuleState(move: Move | Edit, current: number, future: number) {
```
Naviary3: same here

src/client/css/play.css
tsevasa: Oh ups, no idea why it ended up there :D

src/client/scripts/esm/chess/logic/organizedpieces.ts
tsevasa: Sure!

src/client/scripts/esm/chess/logic/state.ts
tsevasa: ok, done!

PR #607, opened 2025-07-18
Update identifiers to use `MemberInfo`

Idonotus:
In preparation for the new `Match` format


src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -81,6 +81,12 @@ interface Variant {
 	}
 }
 
+interface VariantOptions {
+	Variant?: string,
+	UTCDate: string,
+	UTCTime: string
```
Naviary3, line 86: VariantOptions is already the name of a type elsewhere in the project, could this be changed?

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -385,15 +391,15 @@ function getStartingPositionOfVariant(metadata: MetaData) {
  * @param options.Variant - The name of the variant for which to get the gamerules.
  * @returns The gamerules object for the variant.
  */
-function getGameRulesOfVariant(metadata: MetaData): GameRules {
-	if (!isVariantValid(metadata.Variant)) throw new Error(`Cannot get starting position of invalid variant "${metadata.Variant}"!`);
+function getGameRulesOfVariant(variant: VariantOptions): GameRules {
+	if (!isVariantValid(variant.Variant)) throw new Error(`Cannot get starting position of invalid variant "${variant.Variant}"!`);
```
Naviary3: Changing the argument name to variant isn't consistent with the other methods in this script using "metadata"

src/client/scripts/esm/chess/variants/variant.ts
```diff
@@ -385,15 +391,15 @@ function getStartingPositionOfVariant(metadata: MetaData) {
...
 
-	const gameruleModifications: GameRuleModifications = jsutil.deepCopyObject(getVariantGameRuleModifications(metadata));
+	const gameruleModifications: GameRuleModifications = jsutil.deepCopyObject(getVariantGameRuleModifications(variant));
 	
 	return getGameRules(gameruleModifications);
 }
 
-function getVariantGameRuleModifications(metadata: MetaData): GameRuleModifications {
+function getVariantGameRuleModifications(metadata: VariantOptions): GameRuleModifications {
```
Naviary3: If we're changing some variant.ts methods to accept a version of metadata where the UTCDate and UTCTime are required, should we not update all the methods to expect this version?

src/server/game/TypeDefinitions.ts
```diff
@@ -0,0 +1,123 @@
...
+		 * Whether the player was disconnected by choice or not.
+		 * If not, they are given extra time to reconnect.
+		 */
+		wasByChoice?: boolean,
+	};
+}
+
+/** The Game type definition. THIS SHOULD NOT be called, it is purely for JSDoc dropdowns. */
+interface Game {
+	/** The game's unique ID */
```
Naviary3, line 57: Note to self: Also specify that this is the same id that the game will receive in the database

src/server/game/TypeDefinitions.ts
Naviary3, line 1: Almost feel we should disband TypeDefinitions.ts since all of its types are for the server-side Game and might fit better in gamemanager or gameutility. Do you have an opinion?

src/server/game/gamemanager/activeplayers.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Removes the user from the list of users currently in an active game.
+ * This allows them to join a new game.
+ * Doesn't remove them if they are already in a new game of a different ID.
+ * @param user - An object containing either the `member` or `browser` property.
+ * @param gameID - The id of the game they are in.
+ */
+function removeUserFromActiveGame(user: MemberInfo, gameID: number) { // { member/browser }
+	if (!user) return console.error("user must be specified when removing user from players in active games.");
+	if (gameID === undefined) return console.error("gameID must be specified when removing user from players in active games.");
```
Naviary3: Now that this file has type safety we don't need to have these safety checks anymore.

src/server/game/gamemanager/activeplayers.ts
```diff
@@ -0,0 +1,110 @@
...
+function hasColorInGameSeenConclusion(game: Game, color: Player): boolean {
+	const player = game.players[color]; // { member, user_id }  OR  { browser }   (only contains one)
+	if (!player) throw new Error(`Invalid color "${color}" when checking if color in game has seen game conclusion!`);
+
+	return getIDOfGamePlayerIsIn(player.identifier) !== game.id;
+}
+
+/**
+ * Gets a game by player.
+ * @param player - The player object with one of 2 properties: `member` or `browser`, depending on if they are signed in.
```
Naviary3: This argument's jsdoc is innacurate

src/server/game/gamemanager/activeplayers.ts
```diff
@@ -0,0 +1,110 @@
...
+	} else {
+		if (browsersInActiveGames[user.browser_id] === gameID) delete browsersInActiveGames[user.browser_id];
+		else if (browsersInActiveGames[user.browser_id] !== undefined) console.log("Not removing browser from active games because they speedily joined a new game!");
+	}
+}
+
+/**
+ * Returns true if the player behind the socket is already in an
+ * active game, which means they're not allowed to join a new one.
+ * @param {CustomWebSocket} ws - The websocket
```
Naviary3: jsdoc doesn't need duplicate argument type declarations as that's in the function signature (it looks like this was the only one missed here)

src/server/game/gamemanager/afkdisconnect.ts
```diff
@@ -40,13 +41,13 @@ const timeBeforeAutoResignByDisconnectMillis_NotByChoice = 1000 * 60; // 60 seco
  * Cancels the timer that automatically resigns a player due to being AFK (Away From Keyboard).
  * This function should be called when the "AFK-Return" websocket action is received, indicating
  * that the player has returned, OR when a client refreshes the page!
- * @param {Game} game - The game
+ * @param game - The game
  * @param {Object} [options] - Optional parameters.
  * @param {boolean} [options.alertOpponent=false] - Whether to notify the opponent that the player has returned. This will cause their client to cease counting down the time until their opponent is auto-resigned.
```
Naviary3: Missed duplicate argument type declarations

src/server/game/gamemanager/afkdisconnect.ts
```diff
@@ -79,14 +81,13 @@ function startDisconnectTimer(game, color, closureNotByChoice, onAutoResignFunc)
 	// We can do this because if player is disconnected, they are afk anyway.
 	// And if if they reconnect, then they're not afk anymore either.
 	if (game.whosTurn === color && gameutility.isAFKTimerActive(game)) {
```
Naviary3: Okay maybe it wasn't useful for me to add the isAFKTimerActive() method, it's just a one liner, and without it we don't have to use the non null assertion operator after game.autoAFKResignTime below, I rekon we can delete isAFKTimerActive()

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -8,30 +8,33 @@ import WebSocket from 'ws';
...
 import { executeSafely_async } from '../../utility/errorGuard.js';
 import gamelogger from './gamelogger.js';
 
+// @ts-ignore
 import { getTimeServerRestarting } from '../timeServerRestarts.js';
 import { cancelAutoAFKResignTimer, startDisconnectTimer, cancelDisconnectTimers, getDisconnectionForgivenessDuration } from './afkdisconnect.js';
+// @ts-ignore
 import { incrementActiveGameCount, decrementActiveGameCount, printActiveGameCount } from './gamecount.js';
+// @ts-ignore
 import { closeDrawOffer } from './drawoffers.js';
```
Naviary3, line 14: Group non-ts imports together

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -8,30 +8,33 @@ import WebSocket from 'ws';
...
- * Type Definitions
- * @typedef {import('../TypeDefinitions.js').Game} Game
- */
-
-/** @typedef {import("../../socket/socketUtility.js").CustomWebSocket} CustomWebSocket */
-
+import type { Game, PlayerData } from '../TypeDefinitions.js';
+import type { CustomWebSocket } from '../../socket/socketUtility.js';
+// @ts-ignore
+import type { Invite } from '../invitesmanager/inviteutility.js';
```
Naviary3, line 30: Here too (just move it to either the top or bottom since it's the only one)

src/server/socket/socketUtility.ts
```diff
@@ -105,10 +106,8 @@ function getSimplifiedMetadata(ws: CustomWebSocket) {
  * @param ws - The websocket
  * @returns An object that contains either the `member` or `browser` property.
  */
-function getOwnerFromSocket(ws: CustomWebSocket): { member: string, user_id: number } | { browser: string } {
-	const metadata = ws.metadata;
-	if (metadata.memberInfo.signedIn) return { member: metadata.memberInfo.username, user_id: metadata.memberInfo.user_id };
-	else return { browser: metadata.cookies['browser-id']! };
+function getOwnerFromSocket(ws: CustomWebSocket): MemberInfo {
+	return ws.metadata.memberInfo;
```
Naviary3: Now that getOwnerFromSocket() is a one liner I think we can delete it.

src/server/socket/socketUtility.ts
```diff
@@ -162,7 +161,7 @@ function getIPFromWebsocketUpgradeRequest(req: IncomingMessage): string | undefi
  */
 function getSignedInAndIdentifierOfSocket(ws: CustomWebSocket) {
 	const signedIn = ws.metadata.memberInfo.signedIn;
-	const identifier = signedIn ? ws.metadata.memberInfo.username : ws.metadata.cookies['browser-id'];
+	const identifier = signedIn ? ws.metadata.memberInfo.username : ws.metadata.cookies['browser-id']!;
```
Naviary3, line 164: The browser-id cookie should be required, then we can delete all non null assertion operators behind it.

src/server/middleware/verifyJWT.js
Naviary3, line 1: Let's make verifyRefreshToken_WebSocket() also set the browser_id property when they are signed in.

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -172,18 +176,19 @@ function getGameByPlayer(player) {
...
 	const gameID = ws.metadata.subscriptions.game?.id;
 	if (gameID) return getGameByID(gameID); 
     
 	// The socket is not subscribed to any game. Perhaps this is a resync/refresh?
 
 	// Is the client in a game? What's their username/browser-id?
 	const player = socketUtility.getOwnerFromSocket(ws);
-	if (player.member === undefined && player.browser === undefined) return console.error(`Cannot get game by socket when they don't have authentication! We should not have allowed this socket creation. Socket: ${socketUtility.stringifySocketMetadata(ws)}`);
+	// TODO: decide if needed
+	// if (player.member === undefined && player.browser === undefined) throw new Error(`Cannot get game by socket when they don't have authentication! We should not have allowed this socket creation. Socket: ${socketUtility.stringifySocketMetadata(ws)}`);
```
Naviary3: We can delete this, their browser-id property is always gonna be defined

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -340,32 +345,32 @@ function onGameConclusion(game, { dontDecrementActiveGames } = {}) {
...
  * Called when a player in the game loses by disconnection.
  * Sets the gameConclusion, notifies the opponent.
- * @param {Game} game - The game
- * @param {string} colorWon - The color that won by opponent disconnection
+ * @param game - The game
+ * @param colorWon - The color that won by opponent disconnection
  */
-function onPlayerLostByDisconnect(game, colorWon) {
+function onPlayerLostByDisconnect(game: Game, colorWon: Player) {
 	if (!colorWon) return console.log("Cannot lose player by disconnection when colorWon is undefined");
```
Naviary3: This safety check isn't needed anymore with type safety

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -386,10 +391,10 @@ function onPlayerLostByDisconnect(game, colorWon) {
...
  * Sets a 5 second timer to delete the game in case
  * one of them was disconnected when this happened.
- * @param {Game} game - The game
- * @param {string} colorWon - The color that won by opponent abandonment (AFK)
+ * @param game - The game
+ * @param colorWon - The color that won by opponent abandonment (AFK)
  */
-function onPlayerLostByAbandonment(game, colorWon) {
+function onPlayerLostByAbandonment(game: Game, colorWon: Player) {
 	if (!colorWon) return console.log("Cannot lose player by abandonment when colorWon is undefined");
```
Naviary3: same here

src/server/game/gamemanager/gameutility.ts
```diff
@@ -9,16 +9,21 @@
...
 
+// @ts-ignore
 import { getTimeServerRestarting } from '../timeServerRestarts.js';
+// @ts-ignore
 import { doesColorHaveExtendedDrawOffer, getLastDrawOfferPlyOfColor } from './drawoffers.js';
 import timeutil from '../../../client/scripts/esm/util/timeutil.js';
 import typeutil from '../../../client/scripts/esm/chess/util/typeutil.js';
 import variant from '../../../client/scripts/esm/chess/variants/variant.js';
+// @ts-ignore
 import winconutil from '../../../client/scripts/esm/chess/util/winconutil.js';
```
Naviary3: Group non ts imports together

src/server/game/invitesmanager/inviteutility.ts
```diff
@@ -0,0 +1,95 @@
...
+function safelyCopyInvite(invite: Invite): SafeInvite {
+	const inviteDeepCopy = jsutil.deepCopyObject(invite);
+	return makeInviteSafe(inviteDeepCopy);
+}
+
+function memberInfoEq(u1: MemberInfo, u2: MemberInfo): boolean {
+	console.log(u1, u2);
+	// @ts-ignore
+	if (u1.signedIn) return u1.user_id === u2.user_id;
+	// @ts-ignore
```
Naviary3: To avoid the ts-ignores:
```ts
if (u1.signedIn) {
        if (!u2.signedIn) return false;
        return u1.user_id === u2.user_id;
} else if (u2.signedIn) return false; // This ensures if they have the same browser-id, but mi2 is signed in, they are not equal.
else return u1.browser_id === u2.browser_id;
```

src/server/game/TypeDefinitions.ts
```diff
@@ -0,0 +1,123 @@
...
+	 * 
+	 * If they are signed in, their identifier is `{ member: string }`, where member is their username.
+	 * If they are signed out, their identifier is `{ browser: string }`, where browser is their browser-id cookie.
+	 * 
+	 */
+	identifier: MemberInfo;
+	/** Player's socket, if they are connected. */
+	socket?: CustomWebSocket;
+	/** The last move ply this player extended a draw offer, if they have. 0-based, where 0 is the start of the game. */
+	lastOfferPly: number;
```
Naviary3: This needs to be allowed to be undefined. It's unintuitive for lastOfferPly to be 0 if they have never sent a draw offer.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -54,119 +57,116 @@ import { UNCERTAIN_LEADERBOARD_RD } from './ratingcalculation.js';
...
+	const players: Game['players'] = {};
+	// Set the colors
+	const player1 = player1Socket.metadata.memberInfo; // { member/browser }  The invite owner
+	const player2 = socketUtility.getOwnerFromSocket(player2Socket); // { member/browser }  The invite accepter
+	const { playerColors, colorData } = assignWhiteBlackPlayersFromInvite(invite.color, player1, player2);
+	for (const [c, identifier] of Object.entries(colorData)) {
+		players[Number(c) as Player] = {
+			identifier,
+			disconnect: {},
+			lastOfferPly: 0,
```
Naviary3: Then we don't need to set lastOfferPly here

src/server/game/gamemanager/gameutility.ts
```diff
@@ -54,119 +57,116 @@ import { UNCERTAIN_LEADERBOARD_RD } from './ratingcalculation.js';
...
+ * @param game - The game they are a part of.
+ * @param playerSocket - Their websocket.
+ * @param playerColor - What color they are playing in this game. p.NEU
+ * @param options - An object that may contain the option `sendGameInfo`, that when *true* won't send the game information over. Default: *true*
+ * @param options.sendGameInfo
+ * @param options.replyto - The ID of the incoming socket message. This is used for the `replyto` property on our response.
+ */
+function subscribeClientToGame(game: Game, playerSocket: CustomWebSocket, playerColor: Player, { sendGameInfo = true, replyto }: {sendGameInfo?: boolean, replyto?: number} = {}) {
+	if (!playerSocket) return console.error(`Cannot subscribe client to game when they don't have an open socket! ${game.players[playerColor]}`);
 	if (!playerColor) return console.error(`Cannot subscribe client to game without a color!`);
```
Naviary3: We don't need these safety checks anymore with type safety

src/server/game/gamemanager/gameutility.ts
```diff
@@ -240,69 +240,68 @@ function sendGameInfoToPlayer(game, playerSocket, playerColor, replyto) {
 /**
  * Returns the current elo of all players in the game on the leaderboard
  * of the variant being played, or the INFINITY leaderboard if the variant does not have a leaderboard.
- * @param {Game} game 
- * @returns {PlayerGroup<Rating>} - An object containing the rating for non-guest in the game, and whether we are confident in that rating, IF the variant has a leaderboard.
+ * @param game 
+ * @returns An object containing the rating for non-guest in the game, and whether we are confident in that rating, IF the variant has a leaderboard.
  */
-function getRatingDataForGamePlayers(game) {
+function getRatingDataForGamePlayers(game: Game) {
```
Naviary3: Let's explicitly declare the return type so we don't have to study the function, or hover over the signature, to figure it out.
```ts
function getRatingDataForGamePlayers(game: Game): PlayerGroup<Rating> {
```

src/server/game/gamemanager/gameutility.ts
```diff
@@ -330,38 +329,32 @@ function resyncToGame(ws, game, colorPlayingAs, replyToMessageID) {
...
+ * @param game - The game
+ * @param color - The color of the player
+ * @param options - Additional options
+ * @param [options.replyTo] - If specified, the id of the incoming socket message this update will be the reply to
  */
-function sendGameUpdateToColor(game, color, { replyTo } = {}) {
+function sendGameUpdateToColor(game: Game, color: Player, { replyTo }: { replyTo?: number } = {}) {
 	const playerdata = game.players[color];
-	if (playerdata.socket === undefined) return; // Not connected, can't send message
+	if (playerdata?.socket === undefined) return; // Not connected, can't send message
```
Naviary3, line 511: Use the non null assertion operator when declaring playerdata, then we don't need the `?` here.
Even if it was undefined there would still be a crash a couple lines down when we try to access `socket` from it.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -330,38 +329,32 @@ function resyncToGame(ws, game, colorPlayingAs, replyToMessageID) {
...
-/**
- * 
- * @param {Game} game 
- * @param {Player} color 
- * @returns 
- */
-function getGameUpdateMessageContents(game, color) {
-	const messageContents = {
+function getGameUpdateMessageContents(game: Game, color: Player) {
+	const messageContents: any = {
```
Naviary3: This should be of the GameUpdateMessage type. Perhaps now's a good time to move that type into this file?

Naviary3: We can also explicitly declare the return type of the function, then.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -393,36 +386,30 @@ function sendRatingChangeToAllPlayers(game, ratingdata) {
 /**
  * Calculates the json object we send to the client's containing the
  * rating changes from the results of the rated game.
- * @param {RatingData} ratingdata 
- * @returns {PlayerGroup<{ newRating: { value: number, confident: boolean }, change: number }>
  */
-function getRatingChangeMessageContents(ratingdata) {
-	const messageContents = {};
+function getRatingChangeMessageContents(ratingdata: RatingData) {
+	const messageContents: PlayerGroup<{ newRating: { value: number, confident: boolean }, change: number }> = {};
```
Naviary3: This should explicitly return the `PlayerGroup<PlayerRatingChangeInfo>` type.

Also the declaration of messageContents can reference PlayerRatingChangeInfo too.

Let's move that type to this file, too.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -393,36 +386,30 @@ function sendRatingChangeToAllPlayers(game, ratingdata) {
...
 }
 
-/**
- * 
- * @param {Game} game 
- * @param {Player} color
- * @returns {import('../../../client/scripts/esm/game/misc/onlinegame/onlinegamerouter.js').ParticipantState}
- */
-function getParticipantState(game, color) {
+import type { ParticipantState } from '../../../client/scripts/esm/game/misc/onlinegame/onlinegamerouter.js';
```
Naviary3: Let's just move this type into this file.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -393,36 +386,30 @@ function sendRatingChangeToAllPlayers(game, ratingdata) {
...
+
+function getParticipantState(game: Game, color: Player) {
```
Naviary3: Explicity state the return type as `ParticipantState`

src/server/game/gamemanager/gameutility.ts
```diff
@@ -432,15 +419,15 @@ function getParticipantState(game, color) {
 	// Include other relevant stuff if defined...
 
 	if (isAFKTimerActive(game)) {
-		const millisLeftUntilAutoAFKResign = game.autoAFKResignTime - now;
+		const millisLeftUntilAutoAFKResign = game.autoAFKResignTime! - now;
```
Naviary3: isAFKTimerActive() was also probably a bad idea for me to add since it's a one liner. If we delete it then we don't need the non null assertion operator here.

src/server/game/TypeDefinitions.ts
```diff
@@ -0,0 +1,123 @@
...
+	/** Players's current time remaining, in milliseconds, if the game is timed, otherwise undefined. */
+	timer?: number;
+	/** Contains information about this players disconnection and auto resign timer. */
+	disconnect: {
+		/**
+		 * The timeout id of the timer that will START the auto disconnection timer
+		 * This is triggered if their socket unexpectedly closes,
```
Naviary3: For better type safety let's do:

```ts
disconnect: {
    startID?: ReturnType<typeof setTimeout>,
} & ({
    timeoutID: ReturnType<typeof setTimeout>,
    timeToAutoLoss: ReturnType<typeof setTimeout>,
    wasByChoice: ReturnType<typeof setTimeout>,
} | {
    timeoutID: undefined,
    timeToAutoLoss: undefined,
    wasByChoice: undefined,
})
```
(but retain the jsdoc comments)

src/server/game/gamemanager/gameutility.ts
```diff
@@ -449,11 +436,11 @@ function getParticipantState(game, color) {
...
  * Tests if the given socket belongs in the game. If so, it returns the color they are.
- * @param {Game} game - The game
- * @param {CustomWebSocket} ws - The websocket
- * @returns {string | false} The color they are, if they belong, otherwise *false*.
+ * @param game - The game
+ * @param ws - The websocket
+ * @returns The color they are, if they belong, otherwise *undefined*.
  */
-function doesSocketBelongToGame_ReturnColor(game, ws) {
+function doesSocketBelongToGame_ReturnColor(game: Game, ws: CustomWebSocket) {
```
Naviary3: Explicitly declare the return type

src/server/game/gamemanager/gameutility.ts
```diff
@@ -462,28 +449,31 @@ function doesSocketBelongToGame_ReturnColor(game, ws) {
...
  * Tests if the given player belongs in the game. If so, it returns the color they are.
- * @param {Game} game - The game
- * @param {Object} player - The player object with one of 2 properties: `member` or `browser`, depending on if they are signed in.
- * @returns {string | false} The color they are, if they belong, otherwise *false*.
+ * @param game - The game
+ * @param player - The player object with one of 2 properties: `member` or `browser`, depending on if they are signed in.
+ * @returns The color they are, if they belong, otherwise *false*.
  */
-function doesPlayerBelongToGame_ReturnColor(game, player) {
+function doesPlayerBelongToGame_ReturnColor(game: Game, player: MemberInfo) {
```
Naviary3: same here

src/server/game/gamemanager/gameutility.ts
```diff
@@ -462,28 +449,31 @@ function doesSocketBelongToGame_ReturnColor(game, ws) {
...
 	for (const [splayer, data] of Object.entries(game.players)) {
-		const playercolor = Number(splayer);
-		if (player.member && data.identifier.member === player.member) return playercolor;
-		if (player.browser && data.identifier.browser === player.browser) return playercolor;
+		const playercolor = Number(splayer) as Player;
+		if (player.signedIn !== data.identifier.signedIn) continue;
+		// @ts-ignore
+		if (player.signedIn && data.identifier.user_id === player.user_id) return playercolor;
+		// @ts-ignore
+		if (!player.signedIn && data.identifier.browser_id === player.browser_id) return playercolor;
```
Naviary3: Can we use inviteutility.memberInfoEq() to avoid repetition?
Perhaps there's a more suitable script for that function, since it's just comparing memberInfos, which aren't exclusive to invites..

src/server/game/gamemanager/gameutility.ts
```diff
@@ -553,51 +546,51 @@ function isAFKTimerActive(game) {
...
 /**
  * Return the clock values of the game that can be sent to a client or logged.
  * It also includes who's clock is currently counting down, if one is.
  * This also updates the clocks, as the players current time should not be the same as when their turn first started.
- * @param {Game} game - The game
- * @returns {ClockValues}
+ * @param game - The game
  */
-function getGameClockValues(game) {
+function getGameClockValues(game: Game) {
```
Naviary3:
```ts
function getGameClockValues(game: Game): ClockValues {
```

src/server/game/gamemanager/gameutility.ts
```diff
@@ -616,89 +609,90 @@ function getGameClockValues(game) {
...
- * @param {string} color - The color of the player to send the latest move to
+ * @param game - The game
+ * @param color - The color of the player to send the latest move to
  */
-function sendMoveToColor(game, color) {
+function sendMoveToColor(game: Game, color: Player) {
 	if (!(color in game.players)) return logEventsAndPrint(`Color to send move to must be white or black! ${color}`, 'errLog.txt');
     
-	const message = {
+	const message: any = {
```
Naviary3: This should be of the OpponentsMoveMessage type, and let's move that type into this script now.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -616,89 +609,90 @@ function getGameClockValues(game) {
...
  * Returns the color of the player that played that moveIndex within the moves list.
  * Returns error if index -1
- * @param {Game} game
- * @param {number} i - The moveIndex
+ * @param game
+ * @param i - The moveIndex
  * @returns {Player} - The color that played the moveIndex
  */
-function getColorThatPlayedMoveIndex(game, i) {
+function getColorThatPlayedMoveIndex(game: Game, i: number) {
```
Naviary3: This should explicitly return a Player, never undefined.
Instead of throwing an error let's change it's logic to match that of the one in moveutil.ts, which returns the last player if the index is -1.
```ts
// This is the method inside moveutil.ts, for inspiration

/**
 * Returns the color of the player that played the provided index within the moves list.
 */
function getColorThatPlayedMoveIndex(basegame: Game, index: number): Player {
	const turnOrder = basegame.gameRules.turnOrder;
	// If the starting position of the game is in check, then the player very last in the turnOrder is considered the one who *gave* the check.
	if (index === -1) return turnOrder[turnOrder.length - 1]!;
	return turnOrder[index % turnOrder.length]!;
}
```

src/server/game/gamemanager/gameutility.ts
```diff
@@ -709,9 +703,9 @@ function getColorThatPlayedMoveIndex(game, i) {
  * @param {GameRules} gameRules
  * @param {string} condition - The 2nd half of the gameConclusion: checkmate/stalemate/repetition/moverule/insuffmat/allpiecescaptured/royalcapture/allroyalscaptured/resignation/time/aborted/disconnect
```
Naviary3: Duplicate argument types in jsdoc

src/server/middleware/verifyJWT.js
Idonotus: So browser_id is always present? I guess that would work.

src/server/game/TypeDefinitions.ts
Idonotus: I feel TypeDefinitions should be merged with gameutility since it creates Game objects there

Naviary2: That sounds good.

src/server/middleware/verifyJWT.js
Naviary2: Yeah, well as long as they HAVE the cookie, it just doesn't matter whether they're signed in or not.

src/server/game/invitesmanager/acceptinvite.ts
```diff
@@ -8,30 +8,28 @@
 import { logEventsAndPrint } from '../../middleware/logEvents.js';
 
 // Custom imports
-import { isInviteOursByIdentifier } from './inviteutility.js';
+import { memberInfoEq } from './inviteutility.js';
 import socketUtility from '../../socket/socketUtility.js';
 import { createGame } from '../gamemanager/gamemanager.js';
+// @ts-ignore
 import { removeSocketFromInvitesSubs } from './invitessubscribers.js';
+// @ts-ignore
```
Naviary3, line 14: Group non ts imports together

src/server/game/invitesmanager/invitesmanager.ts
```diff
@@ -188,67 +171,67 @@ function deleteInviteByIndex(ws, invite, index, { dontBroadcast, replyto } = {})
...
 //-------------------------------------------------------------------------------------------
 
 /**
  * Returns the first socket subscribed to the invites list that matches the member/browser property.
  * Typically called when you need to inform a player their invite was accepted.
- * @returns {Socket | undefined} - The websocket, if found, otherwise undefined.
+ * @returns The websocket, if found, otherwise undefined.
  */
-function findSocketFromOwner(owner) { // { member/browser }
+function findSocketFromOwner(owner: MemberInfo) { // { member/browser }
```
Naviary3: If this explicitly returns `CustomWebsocket | undefined`, then (see next comment)

src/server/game/invitesmanager/acceptinvite.ts
```diff
@@ -65,11 +63,11 @@ function acceptInvite(ws, messageContents, replyto) { // { id, isPrivate }
 	// Delete the invite accepted.
 	if (deleteInviteByIndex(ws, invite, index, { dontBroadcast: true })) hadPublicInvite = true;
 	// Delete their existing invites
-	if (deleteUsersExistingInvite(signedIn, identifier, { broadCastNewInvites: false })) hadPublicInvite = true;
+	if (deleteUsersExistingInvite(user, { broadCastNewInvites: false })) hadPublicInvite = true;
 
 	// Start the game! Notify both players and tell them they've been subscribed to a game!
 
-	const player1Socket = findSocketFromOwner(invite.owner); // Could be undefined occasionally
+	const player1Socket = findSocketFromOwner(invite.owner) as CustomWebSocket | undefined; // Could be undefined occasionally
```
Naviary3: (from previous comment) then we shouldn't need to cast here.

src/server/game/invitesmanager/acceptinvite.ts
```diff
@@ -85,10 +83,10 @@ function acceptInvite(ws, messageContents, replyto) { // { id, isPrivate }
 
 /**
  * Tests if the provided message contents/body is valid for canceling an invite.
- * @param {*} messageContents - The body of the incoming websocket message. It should look like: `{ id, isPrivate }`
- * @returns {boolean} true if the message contents is valid for the cancellation of an invite
+ * @param messageContents - The body of the incoming websocket message. It should look like: `{ id, isPrivate }`
+ * @returns true if the message contents is valid for the cancellation of an invite
  */
-function verifyMessageContents(messageContents) {
+function verifyMessageContents(messageContents: any) {
```
Naviary3, line 89: After messageContents has been verified to be of and have to expected types, I rekon this function can return messageContents, but in a new type, perhaps called `CreateInviteMessage`, with all expected properties. Then we have type safety in the remaining of acceptInvite() whenever it reads messageContents.

src/server/game/invitesmanager/createinvite.ts
```diff
@@ -11,37 +11,38 @@
 import { logEventsAndPrint } from '../../middleware/logEvents.js';
 
 // Custom imports
+// @ts-ignore
 import clockweb from '../clockweb.js';
 import { existingInviteHasID, userHasInvite, addInvite, IDLengthOfInvites } from './invitesmanager.js';
 import { isSocketInAnActiveGame } from '../gamemanager/activeplayers.js';
+// @ts-ignore
 import { printActiveGameCount } from '../gamemanager/gamecount.js';
+// @ts-ignore
```
Naviary3, line 22: Group non ts imports together

src/server/game/invitesmanager/createinvite.ts
```diff
@@ -76,12 +77,12 @@ async function createInvite(ws, messageContents, replyto) { // invite: { id, own
...
- * @param {*} messageContents - The incoming websocket message contents (separate from route and action)
- * @param {number} replyto - The incoming websocket message ID, to include in the reply
- * @returns {Invite | undefined} The Invite object, or undefined it the message contents were invalid.
+ * @param ws
+ * @param messageContents - The incoming websocket message contents (separate from route and action)
+ * @param replyto - The incoming websocket message ID, to include in the reply
+ * @returns The Invite object, or undefined it the message contents were invalid.
  */
-function getInviteFromWebsocketMessageContents(ws, messageContents, replyto) {
+function getInviteFromWebsocketMessageContents(ws: CustomWebSocket, messageContents: any, replyto: number) {
```
Naviary3: Explicitly declare the return type

src/server/game/invitesmanager/acceptinvite.ts
```diff
@@ -8,30 +8,28 @@
...
- * @param {*} messageContents - The incoming socket message that SHOULD look like: `{ id, isPrivate }`
- * @param {number} replyto - The ID of the incoming socket message. This is used for the `replyto` property on our response.
+ * @param ws - The socket performing this action
+ * @param messageContents - The incoming socket message that SHOULD look like: `{ id, isPrivate }`
+ * @param replyto - The ID of the incoming socket message. This is used for the `replyto` property on our response.
  */
-function acceptInvite(ws, messageContents, replyto) { // { id, isPrivate }
+function acceptInvite(ws: CustomWebSocket, messageContents: any, replyto: number) { // { id, isPrivate }
 
 	if (isSocketInAnActiveGame(ws)) return sendNotify(ws, "server.javascript.ws-already_in_game", { replyto });
```
Naviary3, line 39: It wouldn't let me comment on the right line, but after calling verifyMessageContents() we can gain more safety in the remainder of acceptInvite() by casting messageContents to a new type, AcceptInviteMessage, that has all the expected properties of the message.

src/server/game/invitesmanager/invitesmanager.ts
```diff
@@ -7,18 +7,17 @@
  */
 
 import socketUtility from '../../socket/socketUtility.js';
-import { isInvitePrivate, makeInviteSafe, safelyCopyInvite, isInviteOurs, isInvitePublic, isInviteOursByIdentifier } from './inviteutility.js';
+import { isInvitePrivate, safelyCopyInvite, isInvitePublic, memberInfoEq } from './inviteutility.js';
+// @ts-ignore
 import { getInviteSubscribers, addSocketToInvitesSubs, removeSocketFromInvitesSubs, doesUserHaveActiveConnection } from './invitessubscribers.js';
+// @ts-ignore
```
Naviary3, line 11: Group non ts imports together

src/server/game/invitesmanager/invitesmanager.ts
```diff
@@ -188,67 +171,67 @@ function deleteInviteByIndex(ws, invite, index, { dontBroadcast, replyto } = {})
...
 
 /**
  * Finds an index by ID, and returns an object: `{ invite, index }`, otherwise undefined.
- * @param {number} id - The invite ID
- * @returns {Object | undefined} An object: `{ invite, index }`, or undefined if the invite wasn't found.
+ * @param id - The invite ID
+ * @returns An object: `{ invite, index }`, or undefined if the invite wasn't found.
  */
-function getInviteAndIndexByID(id) {
+function getInviteAndIndexByID(id: string) {
```
Naviary3: Explicitly declare the return type

src/server/game/invitesmanager/invitesmanager.ts
```diff
@@ -188,67 +171,67 @@ function deleteInviteByIndex(ws, invite, index, { dontBroadcast, replyto } = {})
...
-function findSocketFromOwner(owner) { // { member/browser }
+function findSocketFromOwner(owner: MemberInfo) { // { member/browser }
 	// Iterate through all sockets, until you find one that matches the authentication of our invite owner
 	const subscribedClients = getInviteSubscribers(); // { id: ws }
-	if (owner.member) {
+	if (owner.signedIn) {
 		for (const ws of Object.values(subscribedClients)) {
-			if (ws.metadata.memberInfo.username === owner.member) return ws;
+			// @ts-ignore
+			if (ws.metadata.memberInfo.user_id === owner.user_id) return ws;
```
Naviary3: We should be able to use memberInfoEq in this function and avoid the ts-ignores

src/server/game/statlogger.ts
```diff
@@ -27,26 +29,21 @@ const statsPath = path.resolve('database/stats.json');
...
-//     moveCount: {
-//         all: 0,
-//         classical: 0,
-//         month: 0,
-//     }
-// }
+const stats = await readFile('database/stats.json', 'Unable to read stats.json on startup.') as {
+	moveCount: Record<string, number>,
+	gamesPlayed: {
+		byDay: Record<string, number>,
```
Naviary3, line 35: Instead of casting, modify readFile() to accept a generic type parameter, similar to database.get().

src/server/game/statlogger.ts
```diff
@@ -66,25 +63,19 @@ async function logGame(game) {
...
 	if (stats.gamesPlayed.byDay[day] === undefined) stats.gamesPlayed.byDay[day] = 1;
 	else stats.gamesPlayed.byDay[day]++;
 
-
-
-
-	// Atleast 2 moves have been played. Log the game!
-
+	// TODO/FIXME - rewrite the stats format to be safer
+	// @ts-ignore
```
Naviary3, line 84: I don't care so much about statlogger having 100% type safety. It's mostly a depricated script, we shouldn't rely on it for accurate statistics, but it's still kind of nice to have.

src/server/utility/lockFile.ts
Naviary3, line 1: Don't the methods in here still having their `errorString` arguments defeat the purpose of them now throwing errors if they encounter one? If they are low level functions, let them throw (don't intermittently catch them and re-throw), and let the higher level functions which call these methods catch the errors and log with more context (such as the context they are now passing into the `errorString` argument). Similar to how the methods in refreshTokenManager.ts throw.

src/server/game/updateServerRestart.ts
```diff
@@ -12,10 +12,14 @@ import path from 'path';
...
 
 // Custom imports
+// @ts-ignore
 import { broadCastGameRestarting } from './gamemanager/gamemanager.js';
+// @ts-ignore
 import { writeFile_ensureDirectory } from '../utility/fileUtils.js';
+// @ts-ignore
 import { cancelServerRestart, setTimeServerRestarting } from './timeServerRestarts.js';
 
 /** @typedef {import("../socket/socketUtility.js").CustomWebSocket} CustomWebSocket */
```
Naviary3: Convert this to a ts import

src/server/game/updateServerRestart.ts
```diff
@@ -39,13 +43,18 @@ const allowinvitesPath = path.resolve('database/allowinvites.json');
...
+}
+
 /**
  * The allowinvites.json file in the "database". This needs to periodically be re-read
  * in order to see our changes made to it. This is typcailly
  * done when a new invite is attempted to be created.
  * `{ allowinvites: true, restartIn: false }`
  */
-let allowinvites = await readFile(allowinvitesPath, 'Unable to read allowinvites.json on startup.');
+let allowinvites = await readFile(allowinvitesPath, 'Unable to read allowinvites.json on startup.') as AllowInvites;
```
Naviary3: Now we shouldn't need to cast here but pass in AllowInvites as the generic type parameter.

src/server/game/updateServerRestart.ts
```diff
@@ -88,16 +97,15 @@ const updateAllowInvites = (function() {
...
 		// If this is not called with 'await', it returns a promise.
-		const newAllowInvitesValue = await readFile(allowinvitesPath, `Error locking & reading allowinvites.json after receiving a created invite!`);
-
-		timeLastReadAllowInvites = Date.now();
-    
-		if (newAllowInvitesValue === undefined) { // Not defined, error in reading. Probably file is locked
+		try {
+			allowinvites = await readFile(allowinvitesPath, `Error locking & reading allowinvites.json after receiving a created invite!`) as AllowInvites;
+		} catch (e) {
+			if (e instanceof Error) logEventsAndPrint(e.message, 'errLog.txt');
```
Naviary3:
```ts
const errorMessage = e instanceof Error ? e.message : String(e);
```
This is used consistently elsewhere within the repo and ensures that something is still logged if for some reason the event isn't of the Error instance.

src/server/middleware/banned.ts
```diff
@@ -15,21 +17,24 @@ const bannedPath = path.resolve('database/banned.json');
...
 	console.log("Generated banned file");
-})();
-
-const bannedJSON = await readFile(bannedPath, 'Unable to read banned.json on startup.');
+}
 
+const bannedJSON = await readFile(bannedPath, 'Unable to read banned.json on startup.') as {
+	IPs: Record<string, any>,
+	emails: Record<string, any>,
+	'browser-ids': Record<string, any>
```
Naviary3: We shouldn't need to cast here anymore either, but pass in the type as a generic type parameter.

src/server/middleware/verifyJWT.js
```diff
@@ -61,7 +63,7 @@ function verifyAccessToken(req, res) {
 	console.log("A valid access token was used! :D :D");
 
 	const { user_id, username, roles, allowed_actions } = result;
-	req.memberInfo = { signedIn: true, user_id, username, roles, allowed_actions }; // Username was our payload when we generated the access token
+	req.memberInfo = {...req.memberInfo, signedIn: true, user_id, username, roles, allowed_actions }; // Username was our payload when we generated the access token
```
Naviary3: Suppose it's simpler to replace these two lines with:

```ts
req.memberInfo = {...req.memberInfo, signedIn: true, ...result };
```

src/server/middleware/verifyJWT.js
```diff
@@ -134,7 +136,7 @@ function verifyRefreshToken_WebSocket(ws) {
 	}
 
 	const { user_id, username, roles } = result;
-	ws.metadata.memberInfo = { signedIn: true, user_id, username, roles };
+	ws.metadata.memberInfo = {...ws.metadata.memberInfo, signedIn: true, user_id, username, roles };
```
Naviary3: Same here
```ts
ws.metadata.memberInfo = {...ws.metadata.memberInfo, signedIn: true, ...result }
```

src/server/socket/socketUtility.ts
```diff
@@ -162,7 +152,7 @@ function getIPFromWebsocketUpgradeRequest(req: IncomingMessage): string | undefi
  */
 function getSignedInAndIdentifierOfSocket(ws: CustomWebSocket) {
```
Naviary3: Is this used anywhere in the repo anymore? If not we can deleted it.

src/types.ts
```diff
@@ -5,13 +5,15 @@ interface AuthenticatedRequest extends Request {
 }
 
 type MemberInfo = {
+	browser_id: string,
```
Naviary3: We can't actually make browser_id required, since it's not guaranteed to be defined for http requests, even though it is when used on websockets.

src/server/game/gamemanager/gamemanager.ts
```diff
@@ -8,30 +8,32 @@ import WebSocket from 'ws';
 
 // Custom imports
 
+// @ts-ignore
 import gameutility from './gameutility.js';
 import socketUtility from '../../socket/socketUtility.js';
 import statlogger from '../statlogger.js';
+// @ts-ignore
```
Naviary3, line 9: Group non-ts imports together

src/server/game/invitesmanager/invitesmanager.ts
Idonotus: cant we just compare browser id? i dont want to do more comparasons than we need to

Naviary2: Why would we ever switch up the logic that compares memberInfos? We need to find out if one user matches the user that performed an action.

If someone performs an action while signed out, then signs in and performs another action, their memberInfo will match browser-ids, but the 2nd action's memberInfo will be signed in. In that situation we don't want those memberInfos to be considered equal, because then it could join your signed-in self into a game when it should have been your browser self that joins the game. That's just one example but I don't want to find out what other bugs could arise from mis-comparing memberInfos.

Here, we could replace the whole `if (owner.signedIn) {} else {}` blocks with one `for (const ws of Object.values(subscribedClients)) {}` loop that checks if each invites list subscribed client is the owner we're looking for.

src/server/game/gamemanager/gameutility.ts
```diff
@@ -163,7 +163,7 @@ interface Game {
  * @param replyto - The ID of the incoming socket message of player 2, accepting the invite. This is used for the `replyto` property on our response.
  * @returns The new game.
  */
-function newGame(invite: Invite, id: number, player1Socket: CustomWebSocket | undefined, player2Socket: CustomWebSocket, replyto: number) {
+function newGame(invite: Invite, id: number, player1Socket: CustomWebSocket, player2Socket: CustomWebSocket, replyto: number) {
```
Naviary2: player1Socket isn't guaranteed to be defined. Take this scenario:

Player A creates an invite. Player A loses internet connection, socket closes from network interruption. Server gives Player A a 5-second forgiveness window to reconnect, or their invite will be deleted. Within that 5 seconds, Player B accepts player A's invite. We go through the whole createGame() and newGame() route, even though player A hasn't reconnected yet, so their socket is undefined since they don't have one.

PR #688, opened 2025-10-28
Finalize Board Editor

tsevasa:


src/client/css/play.css
```diff
@@ -20,6 +20,15 @@ html {
     --vh-sub-nav: calc(100vh - var(--nav-bar-height));
     /* The minimum between the viewport width and height */
     --vwh: min(var(--vh-sub-nav), var(--vw-capped));
+
+    /* Base editor piece button widths, scale with viewport but stay within reasonable min/max */
+    --editor-piecebutton-width: clamp(30px, calc(var(--vw-capped) / 20), 45px);
+
+    /* Dynamic font size for editor buttons/tools */
+    --editor-font-size: clamp(0.6rem, calc(var(--vw-capped) / 65), 0.9rem);
```
Naviary3: So we don't have to make multiple elements inside the base editor have dynamic sizes, we could give the entire editor a dynamic font size depending on the screen dimensions, then everything inside it, such as piecebutton widths, section font sizes, can all be set by referencing their parent font size, em. For example, 0.5em would set it to 50% the font size of its parent, then it changes automatically if the parent size changes.

tsevasa: For some reason I had trouble with this, but I now pushed a commit that gets rid of --editor-font-size and bases everything on --editor-piecebutton-width instead.

src/shared/chess/util/typeutil.ts
```diff
@@ -217,6 +217,7 @@ export default {
 	royals,
 	SVGLESS_TYPES,
 	strcolors,
+	strtypes,
```
Naviary3: getRawTypeStr() serves the same purpose, and is exported, could we reuse that instead?

tsevasa: Ups, yes we can!

PR #691, opened 2025-10-30
Add functionality to "Start local game from board editor" button

tsevasa:


src/client/scripts/esm/game/chess/gameloader.ts
```diff
@@ -324,6 +324,50 @@ async function startBoardEditor(): Promise<void> {
 	boardeditor.initBoardEditor();
 }
 
+/** Initializes a custom local game from the board editor. */
+async function startLocalGameFromBoardEditor(options: {
```
Naviary2: gameloader.ts shouldn't have any context of who is making to call to start a local game. Maybe make the name more generic? i.e. startCustomLocalGame()

src/client/scripts/esm/game/chess/gameloader.ts
```diff
@@ -324,6 +324,50 @@ async function startBoardEditor(): Promise<void> {
...
+	additional: {
+		variantOptions: VariantOptions,
+	}
+}): Promise<void> {
+	if (typeOfGameWeAreIn !== 'editor') throw Error("Can't start a local game from a custom position when we are not in the board editor.");
+
+	const viewWhitePerspective = gameslot.isLoadedGameViewingWhitePerspective(); // Retain the same perspective as the board editor
```
Naviary2: I think for consistency, local games should always start from white's perspective, otherwise users may be surprised, even if we do add a rotate board feature in the editor one day.

src/client/scripts/esm/game/chess/gameloader.ts
```diff
@@ -324,6 +324,50 @@ async function startBoardEditor(): Promise<void> {
...
+
+	unloadGame();
```
Naviary2: Then boardeditor.ts can be the one to unload the game instead of gameloader.ts, as gameloader no longer needs to know any state from the current loaded game to start a custom local game.

tsevasa: Implemented.

src/client/scripts/esm/game/chess/gameloader.ts
tsevasa: Implemented.

src/client/scripts/esm/game/chess/gameloader.ts
tsevasa: Implemented.

PR #694, opened 2025-11-01
Board Editor game rules: add global special rights controls

tsevasa:


src/client/scripts/esm/game/boardeditor/egamerules.ts
```diff
@@ -158,6 +203,99 @@ function updateGamerulesGUIinfo(new_gamerulesGUIinfo: GameRulesGUIinfo): void {
...
+	}
+
+	guigamerules.setGameRules(gamerulesGUIinfo); // Update the game rules GUI
+}
+
+/**
+ * When a special rights change gets queued, this function gets called
+ * to potentially set gamerulesGUIinfo.pawnDoublePush and gamerulesGUIinfo.castlingWithRooks to indeterminate
+ * */
+function updateGamerulesUponQueueToggleSpecialRight(gamefile: FullGame, coords: Coords, future: boolean): void {
```
Naviary2: getTypeFromCoords() is a bit more optimal since we only need to know the type.

src/client/scripts/esm/game/boardeditor/eactions.ts
```diff
@@ -274,7 +282,9 @@ async function loadFromLongformat(longformOut: LongFormatIn): Promise<void> {
 	boardeditor.addEditToHistory(edit);
 	annotations.onGameUnload(); // Clear all annotations, as when a game is unloaded
 
-	egamerules.setGamerulesGUIinfo(longformOut.gameRules, stateGlobal); // Set gamerules object according to pasted game
+	// Set gamerules object according to pasted game
+	// Currently, we do not compute and pass { pawnDoublePush, castlingWithRooks } here as it might be unnecessarily expensive to compute this when pasting a game
+	egamerules.setGamerulesGUIinfo(longformOut.gameRules, stateGlobal);
```
Naviary2: Okay that's chill, and a nice touch! Later I think we could adopt reading pasted positions to see if they meet the constraints for a bi-state checkbox, and update that accordingly, skipping those checks when the position is large enough to inherit noticebly more load time.

src/client/scripts/esm/game/boardeditor/egamerules.ts
tsevasa: Good point, done!

src/client/scripts/esm/game/boardeditor/eactions.ts
tsevasa: True, sounds good!

src/client/scripts/esm/game/boardeditor/egamerules.ts
```diff
@@ -150,6 +174,38 @@ function setGamerulesGUIinfo(gameRules: GameRules, state_global: Partial<GlobalG
...
+
+	updateGamefileProperties(undefined, undefined, "white");
+	guigamerules.setGameRules(gamerulesGUIinfo); // Update the game rules GUI
+}
+
+/**
+ * This gets called when undoing or redoing moves, to forget the pawnDoublePush and castlingWithRooks entries of the gamerules
+ * since we do not keep track of the checkbox state between edits
+ */
+function scramblePositionDependentGameRules(): void {
```
Naviary2: Okay, understandable. I guess the alternative is, on every applied edit, search it for all specialrights changes and see if they should affect the checkbox states. That could slow down making large edits.

Naviary2: Then again we could do it if the edit isn't too big.

src/client/scripts/esm/game/boardeditor/boardeditor.ts
```diff
@@ -267,6 +269,7 @@ function queueAddPiece(gamefile: FullGame, edit: Edit, coords: Coords, type: num
 	const piece: Piece = { type, coords, index: -1 };
 	boardchanges.queueAddPiece(edit.changes, piece);
 	if (specialright) queueSpecialRights(gamefile, edit, coords, true);
+	egamerules.updateGamerulesUponQueueAddPiece(type, specialright);
```
Naviary2: This method isn't needed if we the placer adds pieces with/without specialrights, to match the "Global special rights" rules.

Naviary2: For example, placing rooks would automatically give them specialrights if "Castling" is checked.

tsevasa: Ah, I see what you mean... Ok, that sounds like a good idea.

src/client/scripts/esm/game/boardeditor/egamerules.ts
```diff
@@ -150,14 +174,123 @@ function setGamerulesGUIinfo(gameRules: GameRules, state_global: Partial<GlobalG
...
+	if (gamerulesGUIinfo.castlingWithRooks !== undefined) {
+		const type = boardutil.getTypeFromCoords(gamefile.boardsim.pieces, coords);
+		if (type !== undefined) {
+			const rawtype = typeutil.getRawType(type);
+			if (castlingWithRooksTypes.includes(rawtype)) {
+				if (gamerulesGUIinfo.castlingWithRooks !== future) gamerulesGUIinfo.castlingWithRooks = undefined;
+			}
+			else if (!pawnDoublePushTypes.includes(rawtype)) {
+				if (future) gamerulesGUIinfo.castlingWithRooks = undefined;
+			}
```
Naviary2: Oh, queueToggleSpecialRight() which calls this function already has access to the Piece, so we don't even need to look this up but could pass it as an argument! 😄 We can replace the coords argument with type.

Naviary2: I don't know why my comment appears to have been placed on different lines than I put it in vscode, but it focuses this line in updateGamerulesUponQueueToggleSpecialRight():
```ts
const type = boardutil.getTypeFromCoords(gamefile.boardsim.pieces, coords);
```

tsevasa: Good catch, done!

src/client/scripts/esm/game/boardeditor/egamerules.ts
tsevasa: Implemented now, by keeping track of these rule states in the edit history.

src/client/scripts/esm/game/boardeditor/boardeditor.ts
```diff
@@ -264,12 +264,19 @@ function queueRemovePiece(gamefile: FullGame, edit: Edit, piece: Piece): void {
...
+ */
+function queueAddPiece(gamefile: FullGame, edit: Edit, coords: Coords, type: number, specialright: boolean | undefined): void {
 	const piece: Piece = { type, coords, index: -1 };
 	boardchanges.queueAddPiece(edit.changes, piece);
 	if (specialright) queueSpecialRights(gamefile, edit, coords, true);
-	egamerules.updateGamerulesUponQueueAddPiece(type, specialright);
+	else if (specialright === undefined) {
+		const { pawnDoublePush, castlingWithRooks } = egamerules.getPositionDependentGameRules();
+		if (pawnDoublePush && egamerules.pawnDoublePushTypes.includes(typeutil.getRawType(type))) queueSpecialRights(gamefile, edit, coords, true);
+		else if (castlingWithRooks && egamerules.castlingWithRooksTypes.includes(typeutil.getRawType(type))) queueSpecialRights(gamefile, edit, coords, true);
```
Naviary2: This logic doesn't need to run for every single queueAddPiece() call, since no selection transformation can attach or detach a special right to a piece type that didn't already have it. This should go inside drawingtool.ts before it calls queueAddPiece(), passing in the respective specialright boolean, instead of undefined as it is passing now.

tsevasa: Good idea, done!

PR #730, opened 2025-11-29
Add contribution_count_singular translation

PilkeySEK:
image

Before, this would show `1 Contributions`, which is not correct. I also added the German translation while at it, I *think* the process was correct?


translation/changes.json
```diff
@@ -1,4 +1,8 @@
 {
+	"76": {
+		"note": "Added contribution_count_singular and renamed contribution_count to contribution_count_plural.",
```
Naviary2: Can you include the 2 line numbers in the change note?

Naviary2:
Thanks!

PR #738, opened 2025-11-30
Change a lot of username/password/email validation to use shared logic

### LINKED ISSUE #732: Clarify what a "valid password" is
Image

It's unclear what this means, so it should be clarified like "Only \<whatever chars> are allowed"
```ts
/^[a-zA-Z0-9!@#$%^&*?]+$/
```

PilkeySEK:
I'm gonna try my best to summarize the changes in detail:
- `src/shared/util/validators.ts`: The new shared logic
  - The logic validates the following:
    - Password too short
    - Password too long
    - Password is 'password'
    - Password has a 'invalid format', meaning the password regex `/^[a-zA-Z0-9!@#$%^&*\?]+$/` didn't match (= disallowed chars)
    - Email has a 'invalid format', meaning the email regex didn't match
    - Email too long
    - Username too short
    - Username too long
    - Username does not have only letters and numbers (`/^[a-zA-Z0-9]+$/`)
    - Username is reserved
- `src/client/scripts/esm/util/password-validation.ts`: **deleted** (the file(s) that were using it use the shared logic now too)
- `src/client/scripts/esm/views/createaccount.js`: This file was using the deleted `password-validation.ts`, changed to using the shared logic
  - `validEmail()` and `onlyLettersAndNumbers()` deleted (the shared logic handles this)
  - Used shared logic, with switch statements to check for the errors
- `src/client/scripts/esm/views/resetpassword.ts`: This file was using the deleted `password-validation.ts`, changed to using the shared logic
- `src/server/controllers/createAccountController.ts`: Updated to use the shared logic
  - `doUsernameFormatChecks()` and `doEmailFormatChecks()` renamed to `doEmailValidation()` and `doUsernameValidation()`
  - Removed reserved usernames from here, since it's in the shared logic now
  - Removed `onlyLettersAndNumbers()` and `isValidEmail()` - both are checked by the shared logic
- Added a translation for when the email is too long

As stated in some comments, the validation could be improved a lot by probably adding a class or some other way of doing things like `.getTranslation()` and `.isValid()` instead of checking manually that the result is equal to `Ok`. I didn't want to blow up this PR with that much code though, (as Naviary said), so this would be something that could be added in the future sometime (once this PR is merged there can be another PR for it i guess)

This also closes #732

### DIFF
--- a/translation/changes.json
+++ b/translation/changes.json
@@ -1,18 +1,6 @@
 {
-  "80": {
-    "note": "Remove js-pwd_incorrect_format on line 168.",
-    "changes": []
-  },
-  "79": {
-    "note": "Remove ws-password_format on line 642.",
-    "changes": []
-  },
-  "78": {
-    "note": "Remove js-username_specs.",
-    "changes": []
-  },
   "77": {
-    "note": "Added js-email_too_long on line 159.",
+    "note": "Added js-email_too_long on line 159. Removed js-username_specs on line 155. Removed ws-password_format on line 642. Removed js-pwd_incorrect_format on line 168.",
     "changes": []
   },
   "76": {
--- a/translation/de-DE.toml
+++ b/translation/de-DE.toml
@@ -1,6 +1,6 @@
 name = "Deutsch" # Name of language
 direction = "ltr" # Change to "rtl" for right to left languages
-version = "80"
+version = "77"
 maintainer = "estetique_bs,tsevasa"
 
 [header]
--- a/translation/en-US.toml
+++ b/translation/en-US.toml
@@ -1,6 +1,6 @@
 name = "English" # Name of language
 direction = "ltr" # Change to "rtl" for right to left languages
-version = "80"
+version = "77"
 maintainer = "Naviary"
 
 [header]


src/shared/util/validators.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Shared logic to validate passwords
+ * @param password The password to check
+ * @returns `Ok` if the password is valid, otherwise another member of that enum
+ */
+function validatePassword(password: string): PasswordValidationResult {
+	if (password.length < 6) return PasswordValidationResult.PasswordTooShort;
+	if (password.length > 72) return PasswordValidationResult.PasswordTooLong;
+	// eslint-disable-next-line no-useless-escape
+	const regex = /^[a-zA-Z0-9!@#$%^&*\?]+$/;
```
Naviary2: The question mark doesn't need to be escaped inside character classes.

src/client/scripts/esm/views/createaccount.js
```diff
@@ -48,13 +47,20 @@ element_usernameInput.addEventListener('input', () => { // When username field c
...
 	}
-    
-	if (lengthError && formatError) { // Change error message
-		usernameError.textContent = translations["js-username_specs"];
-	} else if (lengthError) {
-		usernameError.textContent = translations["js-username_tooshort"];
-	} else if (formatError) {
-		usernameError.textContent = translations["js-username_wrongenc"];
+
+	if (error !== validators.UsernameValidationResult.Ok) {
```
Naviary2: This block is the exact same condition check as above, and can be merged with that.

src/client/scripts/esm/views/createaccount.js
```diff
@@ -48,13 +47,20 @@ element_usernameInput.addEventListener('input', () => { // When username field c
...
+		switch (error) {
+			// TODO: add translation for both of the cases, too long and too short
+			case validators.UsernameValidationResult.UsernameTooShort:
+				usernameError.textContent = translations["js-username_tooshort"];
```
Naviary2: I see we're no longer using js-username_specs, that can be deleted from the TOML.

src/client/scripts/esm/views/createaccount.js
```diff
@@ -95,10 +101,10 @@ element_emailInput.addEventListener('input', () => { // When email field changes
 
 	let emailError = document.getElementById("emailerror"); // Does an error already exist?
 
-	const error = !validEmail(element_emailInput.value);
+	const error = validators.validateEmail(element_emailInput.value); //!validEmail(element_emailInput.value);
```
Naviary2: This comments confusing. It is supposed to show the old code it replaced? It could be deleted

src/client/scripts/esm/views/createaccount.js
```diff
@@ -113,8 +119,18 @@ element_emailInput.addEventListener('input', () => { // When email field changes
 		element_emailInput.removeAttribute('style');
 	}
     
-	if (error) {
-		emailError.textContent = translations["js-email_invalid"];
+	if (error !== validators.EmailValidationResult.Ok) {
```
Naviary2: This block is also the exact same conditional check as above and can be merged.

src/shared/util/validators.ts
```diff
@@ -0,0 +1,130 @@
...
+	UsernameTooLong,
+	OnlyLettersAndNumbers,
+	UsernameIsReserved,
+}
+
+/**
+ * Usernames that are reserved. New members cannot use these are their name.
+ * 
+ * However, the following have been used:
+ * admin
```
Naviary2: I seized the admin account already, this cmt can be updated, could you do that for me?

src/shared/util/validators.ts
PilkeySEK: I mean i just copied the regex from the client or server code... :sob:

src/client/scripts/esm/views/createaccount.js
PilkeySEK: I was thinking to re-add it in the future, the system isn't good enough for that right now though

src/shared/util/validators.ts
Naviary2: That's weird. Where did that backslash come from then? I checked the difference and it literally wasn't there before...

Naviary2: I see, it's in 2 different locations, the one in createAccountController has the extra backslash..
My bad 😅 Let's use the version though without the backslash.

PilkeySEK: i'll try without it and see what happens

PilkeySEK: ok yeah so the regex was only "wrong" on the server

Naviary2: haha

src/server/controllers/createAccountController.ts
```diff
@@ -362,35 +332,30 @@ async function isEmailDNSValid(email: string): Promise<boolean> {
...
+				});
+				return false;
+			case validators.PasswordValidationResult.PasswordIsPassword:
+				res.status(400).json({
+					message: getTranslationForReq('server.javascript.ws-password_password', req),
+				});
+				return false;
+			default:
+				res.status(400).json({
+					message: 'Email is not valid, but the server could not determine why.',
```
shirsakm: This is the password format checker, the `message` should probably be "Password is not valid, [...]", no?

PilkeySEK: oops, forgot about that, seems i made an error while copy pasting

src/client/scripts/esm/views/createaccount.js
```diff
@@ -163,16 +172,24 @@ element_passwordInput.addEventListener('input', () => {
...
 			passwordError = createErrorElement('passworderror', 'password-input-line');
 			element_passwordInput.style.outline = 'solid 1px red';
 		}
-		// Use the error key from the result to get the correct translation
-		passwordError.textContent = translations[validationResult.errorKey];
+		// I found it to be less bloated by using a Map here, but ideally the validator should already return a translation string to be used
+		/**
+		 * @type Map<validators["PasswordValidationResult"], string>
+		 */
+		const errorKeys = new Map();
```
shirsakm: Consider declaring `errorKeys` once at a module level, instead of declaring it upon every keystroke input.

src/shared/util/validators.ts
```diff
@@ -0,0 +1,125 @@
...
+ * This has shared validators between client and server,
+ * to avoid repeating email/password/username validation
+ * and possibly missing to update things both in client and server
+ *
+ * TODO:
+ * - Return list of errors instead of only one, also removes the need for the `Ok` value
+ * - Possibly return a class (?) with a .getTranslationKey() function or add some other way to do that (then there could also be the .isValid property)
+ */
+
+// i have no idea why eslint complains about unused vars here, maybe someone else knows why and can fix it
```
shirsakm: This is likely because ESLint acts stupid when it comes to TypeScript constructs, and doesn't really understand it. This should ideally be fixed by disabling ESLint's `no-unused-vars` and using it from the `typescript-eslint` instead, for TypeScript files.

Naviary2: I fixed this

src/client/scripts/esm/views/createaccount.js
PilkeySEK: should be fixed

src/shared/util/validators.ts
```diff
@@ -0,0 +1,172 @@
...
+const passwordErrorTranslations = new Map<number, PasswordValidationResultTranslations>();
+passwordErrorTranslations.set(PasswordValidationResult.PasswordTooShort, 'js-pwd_too_short');
+passwordErrorTranslations.set(PasswordValidationResult.PasswordTooLong, 'js-pwd_too_long');
+passwordErrorTranslations.set(PasswordValidationResult.PasswordIsPassword, 'js-pwd_not_pwd');
+
+const emailErrorTranslations = new Map<number, EmailValidationResultTranslations>();
+emailErrorTranslations.set(EmailValidationResult.EmailTooLong, 'js-email_too_long');
+emailErrorTranslations.set(EmailValidationResult.InvalidFormat, 'js-email_invalid');
+
+const usernameErrorTranslations = new Map();
```
shirsakm: Missed the type declaration on the `Map()`.

src/client/scripts/esm/views/resetpassword.ts
```diff
@@ -84,9 +84,17 @@ function initializeForm(elements: FormElements): void {
...
-		const validationResult = validatePassword(password);
-		if (!validationResult.isValid && validationResult.errorKey) {
-			messageElement = createErrorMessageElement(translations[validationResult.errorKey]);
+		const validationResult = validators.validatePassword(password);
+		if (validationResult !== validators.PasswordValidationResult.Ok) {
+			// I found it to be less bloated by using a Map here, but ideally the validator should already return a translation string to be used
+			const errorKeys: Map<number, string> = new Map();
+			errorKeys.set(validators.PasswordValidationResult.PasswordTooShort, 'js-pwd_too_short');
+			errorKeys.set(validators.PasswordValidationResult.PasswordTooLong, 'js-pwd_too_long');
+			errorKeys.set(validators.PasswordValidationResult.PasswordIsPassword, 'js-pwd_not_pwd');
```
shirsakm: I might be wrong, but doesn't the new `getPasswordErrorTranslations()` function have the same utility?

PilkeySEK: oops forgot to update that one

PR #739, opened 2025-11-30
Add prettier & eslint

### LINKED ISSUE #729: Add a formatter and a linter as npm scripts
For the formatter, I'd recommend using prettier, it's just the most popular i guess. And of course eslint as the linter since that's being used currently anyway, just not as a npm script.

PilkeySEK:
Closes #729 

Added prettier and eslint as npm scripts. I fixed all eslint errors, but didn't format it, because I don't think Naviary wants to check almost 400 files for potential malicious code.

Another thing that could be added is a pre-commit hook: https://prettier.io/docs/install#git-hooks, but up to you

## Before merging
Please run `npm run format` and commit the result to here, since I didn't format due to the reasons stated above.

### DIFF
--- a/src/client/scripts/esm/game/chess/gamecompressor.test.ts
+++ b/src/client/scripts/esm/game/chess/gamecompressor.test.ts
@@ -1,4 +1,3 @@
-
 import { describe, it, expect } from 'vitest';
 import gamecompressor from './gamecompressor.js';
 import type { FullGame } from '../../../../../shared/chess/logic/gamefile.js';
@@ -17,7 +16,7 @@ describe('gamecompressor', () => {
 				UTCTime: '12:00:00',
 				Variant: 'standard',
 				White: 'Rick',
-				Black: 'Waterman'
+				Black: 'Waterman',
 			} as const;
 
 			const mockGame: FullGame = {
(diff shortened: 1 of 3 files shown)


PilkeySEK:
FirePlank pointed out that I could've just deleted unused variables instead of disabling the eslint warning for it, I wasn't sure though.
You can search for `eslint-disable-next-line no-unused-vars` and look at each one and decide whether to delete them or not i guess

Naviary2:
> FirePlank pointed out that I could've just deleted unused variables instead of disabling the eslint warning for it, I wasn't sure though.
> You can search for `eslint-disable-next-line no-unused-vars` and look at each one and decide whether to delete them or not i guess

I just recently went through and cleared out all unused variables. The exceptions I wanted to keep are preceded with `_` which tells eslint to ignore it. Where did your errors come from??

eslint.config.js
```diff
@@ -81,4 +79,5 @@ export default [
 			]
 		},
 	},
+	eslintConfigPrettier,
```
Naviary2, line 82: Adding this means the lint workflow no longer enforces PRs code to have required styles. What do we do about that? We can't expect all contributors to have preprettified their changes.

(see the team discord channel for discussion solutions)

PilkeySEK: i guess this is solved

PR #748, opened 2025-12-01
Manual Copy/Paste

Tilted-Development:
Adds some _currently unstyled_ dialogue boxes for copying and pasting if a browser doesn't support/permit it.

### DIFF
--- a/translation/en-US.toml
+++ b/translation/en-US.toml
@@ -433,8 +433,8 @@ copied_position = "Copied position to clipboard!"
 loaded_position_from_clipboard = "Loaded position from clipboard!"
 reset_position = "Position was reset!"
 clear_position = "Position was cleared!"
-clipboard_manual_paste = "Enter code:"
-clipboard_manual_copy = "Code:"
+clipboard_manual_paste = "Paste text:"
+clipboard_manual_copy = "Copy text:"
 
 [play.javascript.rendering]
 on = "On"


src/client/scripts/esm/game/chess/pastegame.ts
```diff
@@ -92,8 +92,9 @@ async function callbackPaste(_event: Event): Promise<void> {
 	try {
 		clipboard = await navigator.clipboard.readText();
 	} catch (error) {
+		clipboard = prompt(translations['copypaste'].clipboard_manual_paste);
```
Naviary2, line 95: prompts can return `null` when "cancel" is clicked. In that scenario let's early exit via `return;`!

translation/en-US.toml
```diff
@@ -110,6 +110,7 @@ code_heading = "Code"
 code_credits = [
 "by Brandon Jones and Colin MacKenzie IV.",
 "by Andreas Tsevas and Naviary.",
+"by Tilted Development LTD.",
```
Naviary2, line 113: This section is only intended for giving credit for external libraries used!

translation/en-US.toml
```diff
@@ -432,6 +433,8 @@ copied_position = "Copied position to clipboard!"
 loaded_position_from_clipboard = "Loaded position from clipboard!"
 reset_position = "Position was reset!"
 clear_position = "Position was cleared!"
+clipboard_manual_paste = "Paste text:"
+clipboard_manual_copy = "Copy text:"
```
Naviary2, line 437: These feel more appropriate immediately after `clipboard_invalid`, as those keys are referenced in the same logic.

translation/en-US.toml
Naviary2, line 1: Please include an entry in `translation/changes.json` for all modifications to the english TOML, and increment the english TOML version number at the top of the file!

src/client/scripts/esm/util/docutil.ts
```diff
@@ -34,6 +34,7 @@ function copyToClipboard(text: string): void {
 			console.log('Copied to clipboard');
 		})
 		.catch((error) => {
+			prompt(translations[copypaste].clipboard_manual_copy, text)
```
Naviary2, line 37: copypaste should be quoted

PR #752, opened 2025-12-01
Add world border check to insuffmat script

tsevasa:


src/shared/chess/logic/insufficientmaterial.ts
Naviary2, line 1: Can you replace the comments `//` above the newly added variables with jsdoc `/* ... */`? Intellisense doesn't add the comments as a variable description when hovering over references to them otherwise.

tsevasa: Sure, done!

PR #759, opened 2025-12-03
Migrate to SES for email sending

### LINKED ISSUE #754: Fix email server issues
It seems that using gmail to automatically send out account creation and password reset emails for free is no longer an option since gmail started disabling the website's email account. Some reliable alternative needs to be implemented - either a paid email provider supporting our usecase, or a selfhosted email server.

Important features:
- The email server should be rate limited so that no one can DDOS or otherwise abuse it.
- The infiniteches server should not crash if the email server has issues. Ideally, there should be an error displayed to the client if the email server is having issues.

Naviary2:
closes #754

## To Do:

- [x] Refactor sendMail to use SES instead of gmail.
- [x] Update environment variables.
- [x] Find out why the server didn't catch the original smtp error.
- [x] Add the honeypot bot killer to the create account form.
- [x] Install the `express-rate-limit` package.
- [x] Add route-specific rate limiting to the create account, and send email end points.
- [x] Remove setting up email testing in the Navigation Guide
- [x] Get SES Production Access
- [x] Create an email black list table
- [x] Listen for bounced emails from SES and blacklist those emails so we don't resend to them again.

## After merging, inform contribotors to:

* Run `npm i`.
* Delete `.env` so it can be auto-regenerated.
* Any app passwords you were previously using for testing email sending in dev may be deleted now. Email testing in dev now requires contributors create an AWS account and generate an access key. If keys are not specified, account verification and password reset links are still printed to the console.
https://myaccount.google.com/apppasswords

### DIFF
--- a/src/server/controllers/awsWebhook.ts
+++ b/src/server/controllers/awsWebhook.ts
@@ -1,5 +1,9 @@
 // src/controllers/awsWebhook.ts
 
+/**
+ * Controller to handle AWS SNS webhooks for SES bounce and complaint notifications.
+ */
+
 import type { Request, Response } from 'express';
 import MessageValidator from 'sns-validator';
 import { addToBlacklist } from '../database/blacklistManager.js';


Naviary2:
Unable to test the webhook to see if we can receive bounce notifications from AWS SES... but we'll see I guess

PR #778, opened 2025-12-06
Update fr-FR.toml

AwesomePhoenix-0:
lines 1 to 181 from https://github.com/Infinite-Chess/infinitechess.org/blob/main/translation/en-US.toml are translated in french.
I'll probably continue later.


AwesomePhoenix-0:
I kinda forgot but I'll continue later probably.

PR #792, opened 2025-12-10
Pass move rule to engine

FirePlank:


src/client/scripts/esm/game/chess/engines/hydrochess.js
```diff
@@ -274,7 +274,7 @@ function convertGameToRustFormat(gamefile, timing, strengthLevel) {
 	// Move clocks - moveRuleState is the halfmove counter for 50-move rule
 	const halfmove_clock =
 		gamefile.boardsim?.state?.global?.moveRuleState ??
-		gamefile.startSnapshot?.state_global?.moveRuleState ??
+		gamefile.boardsim?.startSnapshot?.state_global?.moveRuleState ??
```
Naviary2, line 277: If hydrochess.js was in typescript, bugs like this are instantly caught 😉
Not a requirement, though.

PR #793, opened 2025-12-10
Convert Engine Js Wrapper to TS

FirePlank:


src/client/scripts/esm/game/chess/engines/hydrochess.ts
Naviary2, line 1: So once you remove some castings there reveals some incorrect references. Better look into those to make sure the engine isn't bugged!

PR #843, opened 2025-12-28
Rewindable engine legal move history

FirePlank:


src/client/scripts/esm/game/misc/enginegame.ts
```diff
@@ -73,6 +73,8 @@ let engineWorker: Worker | undefined;
 let move_gen_debug: boolean = false;
 /** If {@link move_gen_debug} is true, this contains the list of moves generated by the engine for the current position, in the form "x,y>x,y". */
 let moves_generated: string[] = [];
+/** Stores the legal moves generated by the engine for each move index in the game history. */
+const moveHistoryLegalMoves: Map<number, string[]> = new Map();
```
Naviary2, line 75: If we are now storing the legal moves for specific move indices, then we shouldn't need the preexisting `moves_generated` anymore, the render function can just read the moves of the current move index to know which ones it needs to render.

src/client/scripts/esm/game/misc/enginegame.ts
```diff
@@ -184,6 +187,18 @@ function onMovePlayed(): void {
 	checkmatepractice.registerHumanMove(); // inform the checkmatepractice script that the human player has made a move
 	if (gamefile.basegame.gameConclusion) return; // Don't do anything if the game is over
 
+	// Clear the legal moves cache for all future positions since they're no longer valid
```
Naviary2: Completely unnecessary block. Moves can only be played at the front of the game, so there will never be future positions in the first place.

src/client/scripts/esm/game/misc/enginegame.ts
Naviary2, line 1: I get a crash when trying to enable the debug mode before I start a game

Naviary2, line 1: If I enable the debug mode on a ply which is not the front of the game, it fetches and saves the moves for the front of the game to the currently viewed ply.

src/client/scripts/esm/game/chess/movesequence.ts
Naviary2, line 1: Every single change in this file is unnecessary. They only call enginegame.updateLegalMovesDisplay(), which only calls frametracker.onVisualChange(). But every method inside this file either calls that already or has some path that calls it.

src/client/scripts/esm/game/misc/enginegame.ts
```diff
@@ -316,6 +319,16 @@ function makeEngineMove(compactMove: unknown): void {
 	if (move_gen_debug) sendPositionToEngine(true);
 }
 
+/**
+ * Updates the legal moves display based on the current position in move history.
+ * This should be called when navigating through moves.
+ */
+function updateLegalMovesDisplay(): void {
```
Naviary2: This function is a completely unnecessary wrapper for onVisualChange()

PR #865, opened 2026-01-03
Improves checkmate detection with Huygens pieces

FirePlank:
Improves the checkmate detection algorithm to correctly handle positions involving Huygens pieces. This ensures accurate game-over detection in complex scenarios.

The changes include:

- Handling special cases involving Huygens, which move to prime number distances, during checkmate detection
- Adding a function to determine if a Huygen can capture or block an attacker, considering potential pins and blockers
- Optimizing the process of finding blocking squares for Huygens, checking prime distance only


Naviary2:
Closing in favor of #867

PR #872, opened 2026-01-09
Updated "1.8" to "1.9" in en-EN.toml

LPDJulius:
I changed
> The current version release is 1.8 on the [Play](https://www.infinitechess.org/play?lng=en-US) page!

to
> The current version release is 1.9 on the [Play](https://www.infinitechess.org/play?lng=en-US) page!

and I didn't forget to increase the version variable.


Naviary2:
This is so minor, it's super easy to change across all tomls by searching and replacing "1.8" with "1.9". Then we can skip a change note and version increment!

PR #881, opened 2026-01-13
improve game rule accessibility on phones

BF5258:
Scrolling only needs to be disabled when the window is being dragged. Otherwise the bottom of the menu is inaccessible.
On phones vh can include the space taken up by the browser's UI. If this and the height of the header is not accounted for the popup may be too tall to fit. 
dvh changes as the height of the UI changes. If this is undesirable, svh uses the minimum size of the viewport.
### Before
Screenshot_Firefox_Before
### After
Screenshot_Firefox_After

### DIFF
--- a/src/client/css/play.css
+++ b/src/client/css/play.css
@@ -18,6 +18,8 @@ html {
 	--nav-bar-height: 41px; /* 40 + 1 for border */
 	/* The viewport height, subtract the navigation bar height. */
 	--vh-sub-nav: calc(100vh - var(--nav-bar-height));
+	/* The viewport height on phones can change. */
+	--dvh-sub-nav: calc(100dvh - var(--nav-bar-height));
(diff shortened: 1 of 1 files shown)


src/client/css/play.css
```diff
@@ -1421,7 +1421,7 @@ svg.collapse {
 /* Game Rules UI */
 #game-rules {
 	max-width: min(220px, 50vw);
-	max-height: 80vh;
+	max-height: calc(80dvh - var(--header-height));
```
Naviary2: What if this was 100vh subtract the header height? I see no reason why it can't take up as much space as it's available

BF5258: On all devices or just phone?

Naviary2: dvh works if that's what ur asking

PR #886, opened 2026-01-15
Board editor: Functionality of buttons "reset", "clear", "load", "save as", "save"

tsevasa:


src/client/scripts/esm/game/boardeditor/actions/eactions.ts
```diff
@@ -59,54 +77,87 @@ import bounds from '../../../../../shared/util/math/bounds';
...
+/** Prevent overlapping IndexedDB saves (single-flight): is save ongoing */
+let positionSaveInFlight = false;
+/** Prevent overlapping IndexedDB writes (single-flight): is save pending */
+let positionSavePending = false;
+
 // Actions ----------------------------------------------------------------------
 
 /** Resets the board editor position to the Classical position. */
-function reset(): void {
+async function reset(): Promise<void> {
```
Naviary2, line 69: The downside to making the reset() and clearAl() actions completely unload and reload the whole editor is the navigation bar and editor UI flashes out and in in one frame, which is a little jarring. I think it is unnecessary to completely unload the editor for those actions. What if we used those actions existing behavior, but also reset the edit history?

src/client/css/play.css
```diff
@@ -1612,6 +1640,133 @@ svg.collapse {
 	color: #999;
 }
 
+/* Board editor load & save positions UI */
```
Naviary2, line 1657: What drove the invites and practice checkmates styles to be one grey box is there was nothing inside them that was interactible, you either click on the whole thing or you don't. I think a better styling for editor saves would be something similar to the leaderboard, a table, with every other row slightly darkened.

src/client/scripts/esm/game/boardeditor/actions/eactions.ts
tsevasa: Reworked, as discussed.

src/client/css/play.css
tsevasa: Hm, fair. But I also wanna make the editor saves list draggable, like the practice checkmates list. For that, we can/should probably reuse all the logic, right?

tsevasa: Also, I actually planned on making the elements in the saved positions list clickable. In particular, in the "Save position as" window clicking on a position should copy that position's name into the input field. That would make overwriting positions much simpler. What do you think?

Naviary3: > Hm, fair. But I also wanna make the editor saves list draggable, like the practice checkmates list. For that, we can/should probably reuse all the logic, right?

A grid doesn't change that right? Isn't default behavior of mobile devices that you can swipe to scroll, automatically? ~~I forgot why the practice checkmates needed custom logic~~

> Also, I actually planned on making the elements in the saved positions list clickable. In particular, in the "Save position as" window clicking on a position should copy that position's name into the input field. That would make overwriting positions much simpler. What do you think?

I don't know, it makes it feel like double clicking the whole thing would load it. But I think we should restrict that to the actual load button. Then we can double click and select and copy text inside the position rows, which we cannot do for invites.

tsevasa: >  I forgot why the practice checkmates needed custom logic.

I think that was to get the same behavior on PC.

Ok, fair, making the position names selectable would solve this. Let's make it a grid and do it like that then.

Naviary3: > I think that was to get the same behavior on PC.

Yeah, I think personally, the ability to interact with internal contents of the positions trumps being able to drag to scroll on desktop. I mean we always have the scroll wheel.

src/client/scripts/esm/game/boardeditor/eautosave.ts
```diff
@@ -127,14 +120,22 @@ function stopPositionAutosave(): void {
...
 	// Do a final save when leaving the page
 	positionDirty = true;
-	void saveCurrentPositionOnce();
+	void autosaveCurrentPositionOnce();
+}
+
+async function clearAutosave(): Promise<void> {
+	try {
+		await indexeddb.saveItem('editor-autosave', undefined);
+	} catch (err) {
```
Naviary2: Dead function. Called nowhere in the code. Please delete.

src/client/scripts/esm/game/boardeditor/eautosave.ts
```diff
@@ -48,49 +38,52 @@ function markPositionDirty(): void {
...
-		if (variantOptions.position.size === 0) {
-			// Don't save empty position, as loading it is currently not supported
-			await indexeddb.saveItem('editor-autosave', undefined);
-		} else
-			await indexeddb.saveItem('editor-autosave', {
-				variantOptions,
-				pawnDoublePush,
-				castling,
-			});
+		const positionname = boardeditor.getActivePositionName() ?? '';
```
Naviary2: Should should be alright to be undefined, correct?

src/client/scripts/esm/game/boardeditor/actions/eautosave.ts
```diff
@@ -48,49 +38,52 @@ function markPositionDirty(): void {
...
-				castling,
-			});
+		const positionname = boardeditor.getActivePositionName() ?? '';
+		const date = new Date();
+		const pieceCount = variantOptions.position.size;
+
+		await indexeddb.saveItem('editor-autosave', {
+			positionname,
+			date,
+			pieceCount,
```
Naviary2, line 69: Do not save excess items other than the minimum we read when loading the auto save. `date` nor `pieceCount` are read by the loader.

src/client/scripts/esm/game/boardeditor/eactions.ts
```diff
@@ -48,65 +48,199 @@ import selectiontool from './tools/selection/selectiontool';
...
+
+		boardeditor.setActivePositionName(undefined);
+		await gameloader.startBoardEditor();
+	}
 }
 
-/** Clears the entire board editor position. */
-function clearAll(): void {
+/** Saves current position under "positionname". */
+async function save(positionname: string): Promise<void> {
```
Naviary2: With several of it's own top-level types, constants, and variables, we should move `save()` to its own script to not clutter eactions.ts, as otherwise it's not immediately clear what constants belong to what button action.

src/client/scripts/esm/game/boardeditor/actions/eactions.ts
```diff
@@ -48,65 +48,199 @@ import selectiontool from './tools/selection/selectiontool';
...
+				variantOptions,
+			},
+		},
+		false,
+		false,
+	);
+}
+
+/** Loads a position from a savestate. */
+async function load(editorSaveState: EditorSaveState): Promise<void> {
```
Naviary2, line 121: `gameloader.startBoardEditorFromCustomPosition()` isn't expected to throw an error. I think the try catch block inside this function is unnecessary, and it can be synchronious. Did you experience a crash during testing? If so, make sure illegal/crashing positions aren't sent to it in the first place.

src/client/scripts/esm/game/boardeditor/eactions.ts
```diff
@@ -48,65 +48,199 @@ import selectiontool from './tools/selection/selectiontool';
...
+	positionSavePending = false;
+
+	try {
+		const variantOptions = getCurrentPositionInformation();
+		const { pawnDoublePush, castling } = egamerules.getPositionDependentGameRules();
+		const timestamp = Date.now();
+		const pieceCount = variantOptions.position.size;
+
+		// Save full info for loading purposes
+		await indexeddb.saveItem(`editor-save-${positionname}`, {
```
Naviary2: Both `'editor-save-'` and `'editor-saveinfo-'` are magic strings, found typed out exactly in both saving and loading logic. Which means in the future if they are modified in one location but not updated in the others, everything breaks. Pull them out into constants.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+let current_modal_unabridged_key: string | undefined = undefined;
+
+const delete_button_svg = `<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 2400 2400" width="26" height="26"><g fill="#333"><path d="M300 639c0-49 35-88 77-88h267c53-2 100-40 117-97l3-10 12-39c7-24 13-45 21-63 34-74 97-126 170-139 17-3 37-3 60-3h347c22 0 42 0 60 3 72 13 135 65 169 140 8 17 14 40 21 62l12 40 3 10c18 56 74 94 127 96h257c42 0 77 40 77 88s-35 87-77 87H377c-42 0-77-39-77-87Z"/><path d="M1160 2200h80c279 0 418 0 508-89 90-88 100-233 119-524l26-419c10-158 15-236-30-286-45-50-122-50-275-50H812c-153 0-230 0-275 50-45 50-40 128-30 286l26 419c19 290 28 436 119 524 90 90 230 90 508 90Zm-135-981a76 76 0 00-82-70 78 78 0 00-68 86l50 526c4 43 41 75 82 70a78 78 0 00 68-86l-50-526Zm432-70c42 4 72 42 68 86l-50 526a76 76 0 01-82 70 78 78 0 01-68-86l50-526a75 75 0 01 82-70Z" fill-rule="evenodd" clip-rule="evenodd"/></g></svg>`;
+const load_button_svg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="-.5 0 7 7" width="20" height="20"><path fill="#333" fill-rule="evenodd" d="M5.495 2.573 1.5.143C.832-.266 0 .25 0 1.068V5.93c0 .82.832 1.333 1.5.927l3.995-2.43c.673-.41.673-1.445 0-1.855"/></svg>`;
```
Naviary2: Storing html in a typescript file doesn't sit with me. Different coding languages should keep to their respective files.

The trash svg used for the load button is already defined in `play.ejs` inside the element with id `clearall`. Pull it out of there, and defined it at the top of `play.ejs` as a symbol, similar to how `header.ejs` has defined many multi-use svgs as symbols it the top of itself. When this script appends new positions to the list, it should be able to reference the svg in the document.

Do the same for the play svg to remain consistent.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+	headerEl: element_header,
+	closeButtonEl: element_closeButton,
+	onOpen,
+	onClose,
+});
+
+// Utilities----------------------------------------------------------------
+
+function onOpen(): void {
+	closeModal();
```
Naviary2: `closeModal()` in the `open()` method indicates a failure of state resetting, which should be done when the UI is closed. This should be unneeded here. Is there a specific scenario bug introduced without this?

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+	closeSavePositionUIListeners();
+}
+
+/** Gets the current open/close mode of the Load Position UI */
+function getMode(): typeof mode {
+	return mode;
+}
+
+function openModal(
+	mode: typeof modal_mode,
```
Naviary2: The argument should not include the `undefined` type of `modal_mode`.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+const element_modalNoButton = document.getElementById('load-position-modal-no')!;
+const element_modalYesButton = document.getElementById('load-position-modal-yes')!;
+
+/** The current open/close mode of the Load Position UI */
+let mode: 'load' | 'save-as' | undefined = undefined;
+
+/** The current mode of the Confirmation dialog modal */
+let modal_mode: 'load' | 'delete' | 'overwrite_save' | undefined = undefined;
+let current_modal_positionname: string | undefined = undefined;
+let current_modal_key: string | undefined = undefined;
```
Naviary2: Group all modal-related variables into one object for managing them. Then there's no confusing what variables are for what feature of the UI. For example:
```ts
const modal = {
  mode: 'load' | 'delete' ...
  current_positionname: ...
  ...
}
```

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+	const positionname = element_saveAsPositionName.value;
+	if (positionname === '') return;
+	if (positionname.length > eactions.POSITION_NAME_MAX_LENGTH) {
+		console.error(
+			`This should not happen, position name input box is restricted to ${eactions.POSITION_NAME_MAX_LENGTH} chars, you submitted ${positionname.length} chars.`,
+		);
+		return;
+	}
+	const key = `editor-save-${positionname}`;
+	const unabridged_key = key.replace('editor-saveinfo-', 'editor-save-');
```
Naviary2: Isn't this identical to the above? There's is no `'editor-saveinfo-'` string in the `key`. Well, by this point the magic strings should be removed anyway so this should have been updated.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+	const previous_save = await indexeddb.loadItem<EditorSaveState>(key);
+
+	if (previous_save === undefined) {
+		await eactions.save(positionname);
```
Naviary2: Isn't enough to ask if a previous save exists by reading for an abridged save? Let's not waste memory fully loading the save.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+ * Update the saved positions list
+ */
+async function updateSavedPositionListUI(): Promise<void> {
+	unregisterAllPositionButtonListeners(); // unregister position button listeners
+	element_savedPositionsToLoad.replaceChildren(); // empty existing position list
+
+	const keys = await indexeddb.getAllKeys();
+
+	for (const key of keys) {
+		if (!key.startsWith('editor-saveinfo-')) continue;
```
Naviary2: Immediately filtering out all keys that don't start with `'editor-saveinfo-'` by using `.filter()`, vs a `continue;` on the first line of the loop, makes it more clear we are only dealing with all indexeddb keys regarding editor saves.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+
+		const unabridged_key = key.replace('editor-saveinfo-', 'editor-save-');
+		const editorAbridgedSaveState = await indexeddb.loadItem<EditorAbridgedSaveState>(key);
+
+		// Name
+		const name_cell = document.createElement('div');
+		const positionname = editorAbridgedSaveState?.positionname;
```
Naviary2: OK I'm seeing a lot of `?`s after reading properties from `editorAbridgedSaveState`, it's clear we need full type safety for this.

Use `zod` and create a schema for validating the EditorAbridgedSaveState type. Then use zod to call `z.safeParse(editorAbridgedSaveState)` to obtain the save **fully typed**. Take inspiration from `EditorSavesAPI.ts`.

It's fine to import the zod package client-side. esbuild automatically builds it into the product.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+		await indexeddb.deleteItem(current_modal_unabridged_key);
+		await updateSavedPositionListUI();
+	} else if (modal_mode === 'load') {
+		// Load position
+		const editorSaveState = await indexeddb.loadItem<EditorSaveState>(
+			current_modal_unabridged_key,
+		);
+		if (editorSaveState === undefined || editorSaveState.variantOptions === undefined) {
+			console.error(
+				`Saved position ${current_modal_unabridged_key} appears to be corrupted, deleting...`,
```
Naviary2: ... then we can use that schema to validate the entirety of the save here as well, instead of only checking if the `variantOptions` property is missing.

Double and triple check the schema is correct against all possible save structures we save to indexeddb, as otherwise in edge cases this could cause the save to be entirely deleted below if one property is unexpected.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,353 @@
...
+		if (positionname !== undefined) name_cell.textContent = positionname;
+		else {
+			console.error(
+				`Saved position entry ${unabridged_key} does not have a valid positionname entry, deleting...`,
+			);
+			await indexeddb.deleteItem(key);
+			await indexeddb.deleteItem(unabridged_key);
+			continue;
+		}
+		const row = document.createElement('div');
```
Naviary2: Extract the logic out for creating one row element to its own function, with arguments for name, date, etc.

Then within it, include documentation for what a final row looks like in the html document.

src/client/views/play.ejs
```diff
@@ -40,24 +44,30 @@
...
+                        </div>
+                        <div data-action="save-position" id="save-position" class="tooltip-dl" data-tooltip="Save position">
+							<svg class="svg-save-position" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24"><path fill="#333" d="M6.75 3h-1C4.2312 3 3 4.2312 3 5.75v12.5C3 19.7688 4.2312 21 5.75 21H6v-6c0-1.2426 1.0074-2.25 2.25-2.25h7.5c1.2426 0 2.25 1.0074 2.25 2.25v6h.25c1.5188 0 2.75-1.2312 2.75-2.75V8.2855a3.25 3.25 0 0 0-.9519-2.298l-2.0355-2.0356a3.2503 3.2503 0 0 0-2.2626-.9517V7.5c0 1.2426-1.0074 2.25-2.25 2.25H9c-1.2426 0-2.25-1.0074-2.25-2.25V3Z"/><path fill="#333" d="M14.25 3v4.5a.75.75 0 0 1-.75.75H9a.75.75 0 0 1-.75-.75V3h6ZM16.5 21v-6a.75.75 0 0 0-.75-.75h-7.5a.75.75 0 0 0-.75.75v6h9Z"/></svg>
```
Naviary2: What do you think of these two svgs for save-as and save? They go better with the load svg, as that has a mini bookmark projecting of it 🤔

Image

Todo (if we agree upon it, if not we can discuss other options):

- [ ] Download from:
* Save-as: https://www.svgrepo.com/svg/495656/save-add
* *Save: https://www.svgrepo.com/svg/495653/save-2
- [ ] Edit them, brightening their fill and stroke color to be consistent with the other svgs on the board editor UI, that is, `#333` (near-black).
- [ ] Compress them with [Compress or Die](https://compress-or-die.com/svg)

src/client/views/play.ejs
```diff
@@ -520,8 +530,113 @@
...
+                                    <button id="load-position-modal-yes" class="btn btn-primary">Yes</button>
+                                </div>
+                            </div>
+                        </div>
+                    </div>
+
+                    <!-- Reset position UI for Board Editor -->
+                    <div id="reset-position-UI" class="center narrow scrolling floating-window hidden">
+                        <div id="reset-position-UI-header" class="floating-window-header">
+                            <span class="unselectable">Reset position</span>
```
Naviary2: The Reset and Clear position UI html should be reordered according to the order they appear in the editor UI bar. That is, before the Load position html.

src/client/css/play.css
```diff
@@ -79,6 +79,18 @@ button {
...
+	text-align: center;
+	margin: 0.2em 0;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+}
+
+.italic {
+	font-style: italic;
+}
```
Naviary2: Move to a more-general css file, `header.css`. This can be used elsewhere too.

src/client/css/play.css
Naviary2, line 1: Have you examined each and every added rule in `play.css` to ensure it's not bloat added by whatever AI you are using? Have you, with chrome dev tools opened, gone through each new rule added, and disabled them one by one, to see if anything changes visually? If they don't change anything, they can be deleted.

src/client/css/play.css
```diff
@@ -1612,6 +1656,169 @@ svg.collapse {
...
+
+.position-name-row input {
+	flex: 1;
+	min-width: 0;
+}
+
+.btn-save {
+	width: 75px;
+	height: 28px;
+	justify-content: center;
```
Naviary2: This ruleset doesn't include any context that it is for the save buttons in the load position editor UI. Maybe change it to `position-name-row > .btnsave` or something like that? As it is, `'.btn-save'` could mean anything. Or change it to `.position-name-btnsave`.

Same for the proceeding 3 rulesets.

src/client/css/play.css
```diff
@@ -1612,6 +1656,169 @@ svg.collapse {
...
+/* Ensure the load-position window can anchor absolute children */
+#load-position-UI {
+	position: relative; /* does not move it; only creates a positioning context */
+}
+
+/* The overlay covers ONLY the load-position-UI */
+.local-modal-overlay {
+	position: absolute;
+	inset: 0;
+	z-index: 1000;
```
Naviary2, line 1750: Same for this and the proceeding two rulesets, their names don't include context for what they're for. Maybe use `#load-position-UI > .local-modal-overlay`? or change it to `.load-position-modal`?

Also what does the "local" mean in the name?

src/client/css/play.css
Naviary2, line 1: Inside position rows, while I CAN double click to select text (kudos for that), when I initially hover over the text my cursor doesn't change to the vertical bar.

EDIT: Come to realize, this is the case for every floating window. No text shows the text cursor when hovering over.

Please don't bandaid patch this by adding a new rule like `cursor: text`, because that should be the **default behavior**. Find what rule is preventing the cursor from auto changing to the text cursor and refactor the ~~position rows~~ floating windows to not inherit it in the first place.

My cursor DOES however correctly change to the 'pointer' when hovering over the buttons on the right, so that is good!

Naviary2, line 1: When I compare the brightness of the modal popup to the editor UI bar, it looks darkened. Can we make this white?

Image

src/client/scripts/esm/game/boardeditor/eactions.ts
Naviary2: I would further put it in a new directory `actions/`

src/client/scripts/esm/game/boardeditor/actions/eactions.ts
tsevasa: I never experienced a crash, I was just a bit paranoid of savedata corruption and thought it couldn't hurt. But I can remove all try catch blocks.

src/client/views/play.ejs
tsevasa: Personally, I don't really like these as much as the current option. I don't really associate these symbols with "Save" and I don't think I have ever seen them used. To me it looks more like a bookmark symbol at best. (For context, I am a Windows user. Are these Save symbols on a Mac?)

src/client/css/play.css
tsevasa: The "local" meant that this "modal" only covers the specific floating window instead of the whole screen. But you are right, my naming was inconsistent anyways, changed.

src/client/views/play.ejs
Naviary2: I guess you're right, it's the most universally recognizable save symbol.

Did you compress both the svgs before using them?

Where did you find them, and what's the license for them?

tsevasa: Yes, I did compress them.

See https://discord.com/channels/1114425729569017918/1260305506757775420/1463688122037436478 . Basically, I found them at https://www.svgrepo.com/svg/311154/save and https://www.svgrepo.com/svg/311155/save-as, they are under an MIT license.

src/client/css/play.css
Naviary2: My apologies. It looks like this is a problem across the whole page, not just the floating windows.

And it looks like a lot of items intentionally had the `unselectable` class.

src/client/scripts/esm/game/boardeditor/eactions.ts
tsevasa: Ok, done, I made a new folder /actions with the scripts eactions.ts, esave.ts and eautosave.ts

src/client/views/play.ejs
tsevasa: What do you think? Are you in favor of keeping these symbols and crediting Microsoft?

src/client/scripts/esm/game/boardeditor/actions/eautosave.ts
tsevasa: That would require adding a new type only for the editor autosave that excludes these entries, since these entries are required for normal editor saves.

That might be a bit inconvient, I don't see much harm in just saving these entries in the editor autosave for completeness sake, it's not like they cost many resources. What do you think?

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,394 @@
...
+const element_modal = document.getElementById('load-position-modal-overlay')!;
+const element_modalCloseButton = document.getElementById('close-load-position-modal')!;
+const element_modalTitle = document.getElementById('load-position-modal-title')!;
+const element_modalMessage = document.getElementById('load-position-modal-message')!;
+const element_modalNoButton = document.getElementById('load-position-modal-no')!;
+const element_modalYesButton = document.getElementById('load-position-modal-yes')!;
+
+// Constants
+
+const SVG_NS = 'http://www.w3.org/2000/svg'; // Presumably, this belongs to another file, but to which?
```
tsevasa: Presumably, this constant should be defined in a different file. But where?

src/client/views/play.ejs
Naviary3: Yeah that's fine. I was just thinking going forward how are we going to be consistent with crediting & including licenses with svgs. Because, if they are anywhere other than in the svg itself, it will be a nightmare to maintain credit.

src/client/scripts/esm/game/boardeditor/actions/eautosave.ts
Naviary3: Okay yeah I guess that's fine. Looks like it would be annoying to create a separate type that picks and chooses properties from both the abridge and regular save types.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
Naviary3: I think the best place for it would be `style.ts`!

src/client/css/play.css
tsevasa: Yes, I think nothing is redundant now.

src/client/views/play.ejs
tsevasa: Added credits to svg metadata

src/client/css/play.css
```diff
@@ -1436,6 +1452,30 @@ svg.collapse {
...
 	color: #222;
+	--editor-floating-window-btn-primary: #1d8ee4;
+	--editor-floating-window-btn-primary-hover: #0d7ed4;
+	--editor-floating-window-btn-secondary: #e0e0e0;
+	--editor-floating-window-btn-secondary-hover: #d5d5d5;
+}
+
+.scrolling.floating-window {
+	overflow: auto;
+}
```
Naviary3, line 1464: Can you flip the order of the classes in the selector so that `.floating-window` comes first? Then it reads a "floating window that is scrolling" vs a "scrolling that is also a floating window".

Same thing for the 3 rulesets below.

src/client/scripts/esm/game/boardeditor/actions/eactions.ts
```diff
@@ -62,51 +66,95 @@ const PIECE_LIMIT_KEEP_TRACK_OF_GLOBAL_SPECIAL_RIGHTS = 2_000_000;
...
 		Round: '-',
-		UTCDate,
-		UTCTime,
+		UTCDate: timeutil.getCurrentUTCDate(),
+		UTCTime: timeutil.getCurrentUTCTime(),
+	};
+	const gameRules = variant.getBareMinimumGameRules();
+	const position: Map<CoordsKey, number> = new Map();
+	const specialRights: Set<CoordsKey> = new Set();
+	const state_global = { specialRights };
```
Naviary3: Explicitly declare as the `GlobalGameState` type.

src/client/scripts/esm/game/boardeditor/actions/eactions.ts
Naviary3: Update file path

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,394 @@
...
+		await IndexedDB.deleteItem(modal_config.save_key);
+		await updateSavedPositionListUI();
+	} else if (modal_config.mode === 'load') {
+		// Load position
+		const editorSaveStateRaw = await IndexedDB.loadItem(modal_config.save_key);
+		const editorSaveStateParsed = esave.EditorSaveStateSchema.safeParse(editorSaveStateRaw);
+		if (!editorSaveStateParsed.success) {
+			console.error(
+				`Invalid EditorSaveState ${modal_config.save_key} in IndexedDB ${editorSaveStateParsed.error}`,
+			);
```
Naviary3, line 212: Should we delete the save in this case, too? Otherwise the IndexedDB state remains corrupted.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,394 @@
...
+	} else openModal('overwrite_save', positionname, saveinfo_key, save_key);
+}
+
+/**
+ * Gets executed when a "load position" button is clicked
+ */
+async function onLoadButtonClick(
+	positionname: string,
+	saveinfo_key: string,
+	save_key: string,
```
Naviary3: This, along with onDeleteButtonClick() below, are unnecessary wrappers for openModal().

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,394 @@
...
+ *       <svg><use href="#svg-load" /></svg>
+ *     </button>
+ *     <button class="btn saved-position-btn"> <!-- Delete -->
+ *       <svg><use href="#svg-delete" /></svg>
+ *     </button>
+ *   </div>
+ * </div>
+ *
+ * */
+async function appendRowToSavedPositionsElement(saveinfo_key: string): Promise<void> {
```
Naviary3: How about `generateSavedPositionRow(saveinfo_key: string): HTMLDivElement`? Then this function has one responsibility of generating a row, and doesn't need to know any context of what the row is for or where it is appended, that is left up to the caller.

src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -0,0 +1,394 @@
...
+ * Update the saved positions list
+ */
+async function updateSavedPositionListUI(): Promise<void> {
+	unregisterAllPositionButtonListeners(); // unregister position button listeners
+	element_savedPositionsToLoad.replaceChildren(); // empty existing position list
+
+	const keys = await IndexedDB.getAllKeys();
+	for (const key of keys) {
+		// make sure that key is of the type saveinfo_key
+		if (key.startsWith(esave.EDITOR_SAVEINFO_PREFIX))
```
Naviary3: I prefer this slightly more, it makes it immediately clear what keys we're dealing with:
```ts
const keys = (await IndexedDB.getAllKeys()).filter((key) => key.startsWith(esave.EDITOR_SAVEINFO_PREFIX));
for (const key of keys) {
	await appendRowToSavedPositionsElement(key);
}
```

Naviary3 (changes requested):

PR #887, opened 2026-01-16
Half-bounded world borders.

BF5258:
As described by Naviary on discord, world border edges can be set to null to allow infinite movement in that direction.
https://discord.com/channels/1114425729569017918/1260305506757775420/1454488421484593307

### To do
- [x] update ICN reader to allow unbounded edge
- [x] update game rule menu to allow unbounded edge

### DIFF
--- a/src/shared/chess/logic/legalmoves.ts
+++ b/src/shared/chess/logic/legalmoves.ts
@@ -429,7 +429,7 @@ function slide_CalcLegalLimit(
 	const limit = [...slideMoveset] as SlideLimits; // Makes a copy
 
 	// First of all, if we're using a world border, immediately shorten our slide limit to not exceed it.
-	enforceWorldBorderOnSlideLimit(boardsim, worldBorder, limit, coords, step); // Mutating
+	enforceWorldBorderOnSlideLimit(worldBorder, limit, coords, step); // Mutating
 	// else console.error("No world border set, skipping world border slide limit check.");
 
 	// Iterate through all pieces on same line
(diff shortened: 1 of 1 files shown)


src/shared/chess/logic/legalmoves.ts
```diff
@@ -480,15 +480,19 @@ function slide_CalcLegalLimit(
...
 ): void {
 	if (worldBorder === undefined) return; // No world border, skip
 
+	if (bounds.boxContainsSquare(worldBorder, coords)) {
+		throw Error('Piece outside world border!');
+	}
+
 	if (worldBorder.left !== null && step[0]) {
-		const stepsToIntersect = (worldBorder.left - coords[1]) / step[0];
+		const stepsToIntersect = (worldBorder.left - coords[0]) / step[0];
```
Naviary3: Fair, this looks like it's patched it, and it looks simpler than my implementation... Perhaps no geometry.ts changes are needed then!

src/shared/chess/logic/icn/icnconverter.ts
```diff
@@ -263,6 +263,7 @@ const possessive = (() => {
 const countingNumberSource = String.raw`[1-9]\d*`; // 1+   Positive. Disallows leading 0's
 const wholeNumberSource = String.raw`(?:0|[1-9]\d*)`; // 0+   Positive. Disallows leading 0's unless it's 0
 const integerSource = String.raw`(?:0|-?[1-9]\d*)`; // Prevents "-0", or numbers with leading 0's like "000005"
+const integerOrNullSource = String.raw`(?:null|${integerSource})`; // Allows null
```
Naviary3: The placeholder for null/Infinity has to carefully be considered, as we can never change it without breaking past game notations. On the discord in #Board Editor, we're in accord to use underscores `_` as the placeholder for world borders walls at infinite distance:
```ts
const unboundedIntegerSource = String.raw`(?:_|${integerSource})`;
```
then for example, a world border entry in icn could look like `0,10,_,_`

src/client/scripts/esm/game/gui/boardeditor/guigamerules.ts
Naviary3, line 1: For the board editor gamerules popup World Border section, I think leaving any one field empty should be interpreted as null/Infinity, that seems intuitive, as to begin with they're all empty anyway.

src/shared/chess/logic/icn/icnconverter.ts
```diff
@@ -26,7 +26,7 @@ import type { GameRules } from '../../variants/gamerules.js';
 import type { MetaData } from '../../util/metadata.js';
 import type { EnPassant, GlobalGameState } from '../state.js';
 import type { BaseRay } from '../../../util/math/geometry.js';
-import { BoundingBox } from '../../../util/math/bounds.js';
+import { UnboundedRectangle } from '../../../util/math/bounds.js';
```
Naviary3: `import type`

src/shared/chess/logic/icn/icnconverter.ts
```diff
@@ -263,6 +263,7 @@ const possessive = (() => {
 const countingNumberSource = String.raw`[1-9]\d*`; // 1+   Positive. Disallows leading 0's
 const wholeNumberSource = String.raw`(?:0|[1-9]\d*)`; // 0+   Positive. Disallows leading 0's unless it's 0
 const integerSource = String.raw`(?:0|-?[1-9]\d*)`; // Prevents "-0", or numbers with leading 0's like "000005"
+const unboundedIntegerSource = String.raw`(?:_|${integerSource})`; // Allows null
```
Naviary3: `// Allows _ as a placeholder for infinity`

src/shared/chess/logic/legalmoves.ts
```diff
@@ -480,56 +476,24 @@ function slide_CalcLegalLimit(
 /** Modifies the provided slide limit in a single step direction (positive & negative) to not exceed the world border. */
 function enforceWorldBorderOnSlideLimit(
```
Naviary3, line 476: Wait- what happened to the original logic in here, it was so clean! I thought we were in agreeance that your logic was better. While I appreciate you making some geometry.ts and bounds.ts methods to hopefully be used by infinite pieces logic in the future, I think as they are now they are too specific to the use case in enforceWorldBorderOnSlideLimit() to be useful elsewhere. I think [this commit](https://github.com/Infinite-Chess/infinitechess.org/pull/887/commits/9178f43f252a9f69ad5c398d305aa9eef80688f9) should be reverted

src/shared/util/math/bounds.ts
```diff
@@ -24,6 +24,17 @@ interface BoundingBox {
 	top: bigint;
 }
 
+interface UnboundedRectangle {
+	/** The x-coordinate of the left side of the box. */
+	left: bigint | null;
```
Naviary3, line 33: Add jsdoc to `UnboundedRectangle` describing it's a variant of `BoundingBox` but with edges potentially at infinity, where `null` is a placeholder for infinity.

Naviary2:
Looks good now!

PR #898, opened 2026-01-28
Await IndexedDB.deleteItem in tandem

tsevasa:
### Type of Change:

- [ ] New feature
- [ ] Bug fix
- [ ] Refactor (no logic change)
- [x] Minor improvement

`guiloadposition.onModalYesButtonPress()`

### Details of what it does:
See diff

### DIFF
--- a/src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
+++ b/src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
@@ -203,7 +203,7 @@ async function onModalYesButtonPress(): Promise<void> {
 			IndexedDB.deleteItem(modal_config.saveinfo_key),
 			IndexedDB.deleteItem(modal_config.save_key),
 		]);
-		await updateSavedPositionListUI();
+		updateSavedPositionListUI();
 	} else if (modal_config.mode === 'load') {
 		// Load position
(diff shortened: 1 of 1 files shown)


src/client/scripts/esm/game/gui/boardeditor/guiloadposition.ts
```diff
@@ -215,7 +217,7 @@ async function onModalYesButtonPress(): Promise<void> {
 				IndexedDB.deleteItem(modal_config.saveinfo_key),
 				IndexedDB.deleteItem(modal_config.save_key),
 			]);
-			updateSavedPositionListUI();
+			await updateSavedPositionListUI();
```
Naviary2: This is the only weird change about this

Naviary2: Only cause it's inconsistent. Everywhere else doesn't await this. And considering its the last method called here, I don't think we have to await. Neither line 206

tsevasa: You're right, it's inconsistent.

But in line 206, it's not the last method called, because closeModal() gets executed afterwards.

tsevasa: But ok, the awaits are not needed anyways. Removed them.

PR #991, opened 2026-02-13
Never exceed provided wb and update hydrochess wb

FirePlank:

refactor

Chess engine world border handling

### Details of what it does (if it makes css style changes, include screenshots of the before & after):

Made world border never exceed given bounds and updated hydrochess world border to be closer to i64.

### DIFF
--- a/src/client/scripts/esm/game/boardeditor/actions/eactions.ts
+++ b/src/client/scripts/esm/game/boardeditor/actions/eactions.ts
@@ -263,23 +263,34 @@ function startEngineGame(engineUIConfig: EngineUIConfig): void {
 		// Calculate minimum bounding box of all pieces
 		const bb = boardutil.getBoundingBoxOfAllPieces(gameslot.getGamefile()!.boardsim.pieces)!; // Guaranteed defined since above we check if there's > 0 pieces
 
-		const limit = engineWorldBorderDict[currentEngine];
+		/*
+		 * Priority:
+		 * 1. Default distance
+		 * 2. Capped at engine's cap
+		 */
+
+		const worldBorderProperty = engineWorldBorderDict[currentEngine];
+		const cap = hydrochess_card.BORDER_CAP;
 
 		// How far can we extend in each direction before hitting ±limit?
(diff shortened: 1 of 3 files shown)


src/client/scripts/esm/game/boardeditor/actions/eactions.ts
```diff
@@ -274,6 +274,27 @@ function startEngineGame(engineUIConfig: EngineUIConfig): void {
 		};
 	}
 
+	// Ensure world border exists and respects engine coordinate limits
+	{
```
Naviary2: A. We shouldn't be changing (capping) the world border if the user didn't select for the world border to be set automatically. In that case, let the engine's hydrochess card naturally report below that the world border isn't supported.

B. Merge this block with the above block, and make it fair (that logic should be here, not in gamefile.ts). Simple clamping won't do. For example, what if the classical position is pushed upward super far to be adjacent to the max world border? Then the top gets clamped, but not the bottom, giving white a huge space advantage. The border should be equidistant from opposite sides of the position. If you clamp one side, the opposite needs to be shortened by the same amount.

src/client/scripts/esm/game/chess/enginecards/hydrochess_card.ts
```diff
@@ -56,7 +56,8 @@ function isPositionSupported(variantOptions: VariantOptions): SupportedResult {
 	}
 
 	// 2. World border larger than i64 is unsupported.
-	const cap = 1_000_000_000_000_000_000n; // About 10% the max, for cushion
+	const I64_MAX = 2n ** 63n - 1n;
```
Naviary2: Magic number declared in multiple locations. I would put this constant in `jsutil.ts`.

src/shared/chess/logic/gamefile.ts
Naviary2, line 1: gamefile.ts shouldn't be doing the balancing of the world border edges, it should load the exact game that was specified in the arguments. This is a job of eactions.ts. No changes required in this file.

Naviary2 (changes requested):
see discord

PR #1123, opened 2026-06-02
fix: remove pause click listener from correct element to stop leak

### LINKED ISSUE #1122: Pause button click listener is never removed, leaking listeners across navigation open/close cycles
### Description

The navigation bar's pause button accumulates a new `click` listener every time
the navigation UI is opened and closed, because the listener is never removed.

In `guinavigation.ts`, `initListeners_Navigation()` attaches `callback_Pause`
to `element_pause`, but `closeListeners_Navigation()` tries to detach it from
`element_Back` instead. Since `element_Back` never had that listener, the
`removeEventListener` call does nothing, and the listener on `element_pause`
stays behind forever.

The navigation UI is opened and closed many times during a normal session —
on every game start/end, when toggling the board editor, when returning from
the pause menu, and so on. Each cycle leaves one more `callback_Pause` listener
attached.

### Impact

- **Memory leak:** listeners (and their closures) pile up over time and are
  never freed. The more games you play in a single session, the more leak.
- **Duplicate handler firing:** a single click on the pause button ends up
  running `callback_Pause` multiple times — once for every listener that has
  accumulated.

### Steps to reproduce

1. Start a game, then leave it (back to the title menu).
2. Repeat a few times.
3. Open DevTools and inspect the listeners on the `#pause` element — the count
   keeps growing. Clicking pause now triggers the handler once per accumulated
   listener.

### Root cause

`closeListeners_Navigation()` removes the pause listener from the wrong element
(`element_Back` instead of `element_pause`) — a copy/paste slip.

BilgeGates:
Closes #1122 

Bug fix

The in-game navigation bar (`guinavigation.ts`).

### Details of what it does:

`callback_Pause` is added to `element_pause` in `initListeners_Navigation()`,
but `closeListeners_Navigation()` was removing it from `element_Back` instead,
so it was never actually detached.

Because the navigation UI opens and closes repeatedly during a session (every
game start/end, board editor toggle, returning from the pause menu, etc.), a new
pause listener was being added on every cycle and none were ever removed. This
leaked listeners over time and caused a single pause click to fire the handler
multiple times.                                                          
The fix detaches the listener from the correct element — `element_pause` instead
of `element_Back` — which targets the root cause directly. One line changed.

`npm run lint`, `npx tsc --noEmit`, and `npm test` all pass.

### DIFF
--- a/src/client/scripts/esm/game/gui/guinavigation.ts
+++ b/src/client/scripts/esm/game/gui/guinavigation.ts
@@ -319,7 +319,7 @@ function closeListeners_Navigation(): void {
 	element_Annotations.removeEventListener('click', callback_Annotations);
 	element_Erase.removeEventListener('click', callback__Collapse);
 	element_Collapse.removeEventListener('click', callback__Collapse);
-	element_Back.removeEventListener('click', callback_Pause);
+	element_pause.removeEventListener('click', callback_Pause);
 
 	element_CoordsX.removeEventListener('change', callback_CoordsXChange);
 	element_CoordsY.removeEventListener('change', callback_CoordsYChange);


Naviary2:
closes #1122

Naviary2:
Nice notice! It's not as severe as claimed as the listener count doesn't actually blow up, since adding the same exact reference as a callback doesn't add a duplicate, but true, it wasn't being removed when the game page was closed, so this is a nice cleanup.

For the meantime can any PRs be directed onto the `redesign` branch? I am actively working there, and pushing changes here means conflicts with that branch.

Naviary2:
Thanks for following the PR guidelines! I now realize a potential issue with it in that 2 pages of text explaining 1 line of changed code is excessive.. at that point it's quicker to read the code changes to understand the fix instead of reading the description. I need to think of how I should update it.

BilgeGates:
Thanks for the merge and the tip about the redesign branch! I will make sure to target it for any future PRs. Good luck with the redesign!

BilgeGates:
Regarding the PR template, if you're looking for ideas, GitHub actually supports Multiple PR Templates (by creating a .github/PULL_REQUEST_TEMPLATE/ folder). You could have a detailed one for 'Features' and a very short one for 'Minor Bug Fixes'.
Alternatively, adding a simple checkbox like 'Is this a minor <5 lines fix?' that allows skipping the long explanation could work perfectly. Just a thought!

Naviary2:
I decided to delete the template entirely and move the description guidelines into GUIDELINES.md

PR #1127, opened 2026-06-03
fix(navigation): stop the navigation bar from leaking button listeners

### LINKED ISSUE #1126: Navigation bar leaks button listeners when toggled across board editor state changes
### Description

The navigation bar attaches a different set of button listeners depending on whether the board editor is open — the rewind/forward buttons during a normal game, or the undo/redo buttons in the editor. Both `initListeners_Navigation()` and `closeListeners_Navigation()` decide which set to touch by calling `guiboardeditor.isOpen()`.

The issue is that the editor's open state can change between the moment listeners are added and the moment they're removed. When that happens, the close routine removes the wrong set and leaves the originally-attached listeners on their elements. They pile up every time the bar is toggled.

### Steps to reproduce

1. Open the navigation bar during a normal game (attaches the move-button listeners).
2. Open the board editor.
3. Toggle the navigation bar closed (`N` key).
4. The close routine now sees `isOpen() === true`, removes the undo/redo listeners, and leaves the move-button listeners attached.

### Expected

Toggling the navigation bar should never leave stale listeners behind, regardless of whether the board editor was opened or closed in between.

BilgeGates:
Closes #1126 

Bug fix

The in-game navigation bar (guinavigation.ts).

### Details of what it does (if it makes css style changes, include screenshots of the before & after):

The navigation bar attaches either the move (rewind/forward) or edit (undo/redo) button listeners depending on `guiboardeditor.isOpen()`. Both the init and close routines branched on that same live state to decide which set to touch.

The problem: the editor's open state can change between attaching the listeners and removing them. For example, opening the nav bar in a normal game attaches the move-button listeners; then opening the board editor and toggling the bar closed makes the close routine see `isOpen() === true`, so it removes the undo/redo listeners and leaves the move-button listeners attached. They accumulate on every toggle.

The fix records which set was actually attached and removes that exact set on close, regardless of the editor's current state — targeting the root cause rather than re-guessing the state.

No CSS changes. `npm run type-check` and `npm run lint` pass.

### DIFF
--- a/src/client/scripts/esm/game/gui/guinavigation.ts
+++ b/src/client/scripts/esm/game/gui/guinavigation.ts
@@ -77,6 +77,9 @@ const durationToLockRewindAfterMoveForwardingMillis = 750;
 /** Whether the navigation UI is visible (not hidden) */
 let navigationOpen = true;
 
+/** Whether the edit (undo/redo) button listeners were attached instead of the move (rewind/forward) ones. */
+let editButtonsListenersActive = false;
+
 /**
  * Whether the annotations button is enabled.
  * If so, all left click actions are treated as right clicks.
@@ -272,7 +275,8 @@ function initListeners_Navigation(): void {
 	element_CoordsX.addEventListener('change', callback_CoordsXChange);
 	element_CoordsY.addEventListener('change', callback_CoordsYChange);
 
-	if (!guiboardeditor.isOpen()) {
+	editButtonsListenersActive = guiboardeditor.isOpen();
+	if (!editButtonsListenersActive) {
 		element_moveRewind.addEventListener('click', callback_MoveRewind);
 		element_moveRewind.addEventListener('mousedown', callback_MoveRewindMouseDown);
 		element_moveRewind.addEventListener('mouseleave', callback_MoveRewindMouseLeave);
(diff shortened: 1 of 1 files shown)


Naviary2:
Hey this script is eventually going to get deleted when the game page is redesigned, so these changes would never see production even if merged. For that reason I don't want me or you to waste time working on stuff that's going to be deleted.

The current status of it is- The whole backend build pipeline has been prepared so assets can be cached by browsers permanently, the template renderer has been migrated from ejs to nunjucks, server side rendering is fully setup, the whole new translation and localization system is prepared for new pages to be localized once they've been created and their design finalized. Working on actual page redesigns now. So far only the home page and login page are done. Next will be the register page. Relevant files are stack.md (planned tech stack of redesign), todo.md (rough remaining todos of redesign, sequential), design.md (incomplete scratch pad of descriptions of each new page). Any scripts that are imported and used on any file besides the home or login page right now have potential to be deleted/rewritten. I plan to delete them as I go, instead of mass deleting all old stuff right away, this is so that some of them I can just migrate over, without having to rewrite them from scratch. Old/stale stuff is being deleted as I design their relevant page.

BilgeGates:
Understood! I completely respect the redesign process and the decision to close the PR.

By the way, I really love what you're building with this project. I've been playing a few games against the computer, and it's a genuinely interesting challenge—especially for those of us who find standard chess a bit repetitive. It's a great concept!

Good luck with the redesign! If you ever need a hand with testing or any other tasks during the transition, please feel free to reach out. I'd be happy to help. 😊

PR #1144, opened 2026-07-14
Add local engine (vs computer) games, saved like PvP games

FirePlank:
Implements the **Play against computer** flow. Engine games run locally in the browser via the Apeiron wasm worker (no server socket, no server-side move validation), but are recorded server-side with a real game id and persisted to the `games`/`player_games` tables on conclusion — so they get a `/game/:id` page, show up in history, and link to the analysis board exactly like PvP games.

## Server
- **`engine_games` table + manager** — one row per game, created at start, storing the human owner (`user_id`/`browser_id`), color, engine, strength, variant/position, clock, and live moves + clock snapshots. Kept after conclusion as the permanent record of the engine participant + settings.
- **`/api/engine-game` endpoints** — create, owner-only resume state, per-move progress sync, and conclude (logs to `games`/`player_games` as **casual**, no ratings/stats; a zero-move game is deleted rather than stored).
- **`gamePageController`** serves live engine games (owner-only) and resolves the engine participant + dead *guest* owners (engine games retain the browser id, unlike PvP); **`deadgamestate`** names the engine instead of falling back to "(Guest)".
- game ids are unique across `games` **and** `engine_games`; stale unconcluded engine games are purged in the periodic cleanup.

## Client
- Modal **Play against computer** creates the game and navigates to `/game/:id`. The variant dropdown **hides** engine-unsupported variants (rather than erroring on submit), and custom ICN positions are validated against the engine card (piece types, world border, piece limit, promotion lines, etc.).
- Game page loads engine games over HTTP, runs the engine locally, and syncs progress/conclusion back. Resign/abort conclude the game locally; draw-offer, chat, and disconnect UI are omitted.
- The gameplay engine now uses the **served multithreaded build** (Lazy SMP, threads = cores − 1 capped at 4), same as the analysis engine, and `/game/:id` is cross-origin isolated. Strength levels raised to **8**.

## Shared
- Engine metadata (engine dictionary + hydrochess card) moved to `src/shared/chess/engines` so both client and server use one source. The default world-border helper is extracted and shared with the board editor.

## Notes
- `enginegame` no longer imports `checkmatepractice` — the practice page registers observer hooks instead, keeping page-specific code off other pages that host engine games.
- Rebased onto `redesign`: adopts the moved variant-selector component (with the engine-only filter restored via a compound `.hidden` selector) and the refactored engine-bounds helper.
- Passes `type-check` + `lint`. Verified end-to-end over HTTP: create → owner-only SSR → resume → per-move sync → conclude → permanent record with correct result/termination + engine name on the dead/analysis pages.


Naviary2:
1. Regarding how are clocks restored on a refresh- Is the time of the last move stored on the server (whether in memory or in the db) as the epoch of the *user's* (not server's) clock?

2. When it's the engine's turn, and we refresh the page, is the engine's clock restored to what it was at the start of its turn, rather from the epoch it will lose at (different from how the user's clock is restored)?

3. Are the client-side engine scripts conditionally injected into game.njk only when the game is actually an engine game, and not loaded for standard user vs user games?

4. If engines are not given accounts, how is the engine name being restored as your opponent name when you refresh the page, whether the game is alive or dead?

5. Revert the changes that compress ICNs before saving them in the database.

FirePlank:
1. Clock restore uses the player’s browser clock. turnStartTime is the start of the currently ticking turn, not “time of the last move” and comes from client-side Date.now(). The server stores that supplied epoch in engine_games.turn_start_time. last_updated uses server time but is only for stale-game cleanup.

2. On the human’s turn, refresh deducts Date.now() - turnStartTime from the stored clock snapshot. On the engine’s turn, refresh deliberately does not deduct elapsed offline time. The engine restarts its search, and its timeout is rebuilt as Date.now() + storedEngineClock, effectively restoring the clock to the start-of-turn snapshot.

3. In game.njk, the engine worker URL and WASM glue URL are injected only when gamePageData.engineGame exists. Standard games do not fetch or initialize the engine worker/WASM. The ordinary game-page bundle statically imports the small engine-game loader/controller code, only the heavy engine assets are conditionally loaded.

4. Engines do not need accounts. The retained engine_games sidecar row stores engine, strength_level, and the human’s color:
- Live refresh: produceEngineGameStaticState() formats the stored engine and strength into the opponent name.
- Dead game: getEngineParticipant() reads the same retained row and inserts the engine participant into the reconstructed player list.

5. Done.

Naviary2:
# Review — PR #1144 "Engine games on the game page"

**Checks:** `npm run type-check` ✅ `npm run lint` ✅ `npm test` ✅ (7 files, 83 tests)

**Summary.** The client-side work here is good and most of it stays: the wasm worker rework, the
lobby's computer-game flow, the variant filtering, the strength range. The **server-side
architecture needs to be replaced** — engine games should be ordinary live games in the existing
`activeGames` / `live_games` pipeline, not a second parallel pipeline with its own table, endpoints,
per-move sync, caps and rate limiting. That decision (§1) deletes most of the PR's server code and
about two thirds of the individual problems this review would otherwise list. Section 2 is bugs to
fix regardless, §3 is redundancy with systems we already have, §4 is dead code.

---

## 1. Architecture: engine games belong in the live-game pipeline

The PR as is makes the server a dumb recorder: a client POSTs its whole move list after every move, and
POSTs whatever result it likes at the end. That means (a) a second live-game persistence pipeline
duplicating what `live_games` + `live_player_games` already do — same job, same compact-ICN-with-
`[%clk]` move format — and (b) a permanent `games`/`player_games` archive whose contents are
entirely client-asserted, which rules out ever counting engine games in player stats.

Instead: an engine game is an ordinary in-memory `ServerGame` where one participant is an engine
whose moves arrive over the owner's socket. The server owns the clocks, the conclusions, and (for
every variant the engine supports) real move legality validation. Note the validation gate is
position-string length — `POSITION_STRING_THRESHOLD = 2500`, with Obstocean at 2425 the largest
engine-supported variant, and the only exclusions (`Omega_Squared/Cubed/Fourth`, `5D_Chess`) aren't
engine-supported — so **no coupling to the engine's own limits is needed**;
`doesVariantSupportServerValidation` decides it per-variant exactly as it does for PvP, and custom
pasted positions fall back to `applyClientReportedMove` like any PvP custom game.

### 1.1 Delete

`engine_games` as designed, `engineGamesManager.ts`'s update/delete/stats/purge paths,
`EngineGameAPI.ts`, `routes/engineGame.ts`, `enginegamesync.ts`, `enginegameloader.ts`'s fetch/sync
half, the `EngineGameProgressBody` / `ConcludeEngineGameBody` / `EngineGameState` schemas,
`cleanupTasks`' engine purge, `addTurnStartTimeColumnToEngineGamesIfNeeded`,
`engineGameSyncLimiter`, `ENGINE_GAME_MOVES_STRING_CAP`, `ENGINE_GAME_MAX_PLIES`,
`countEngineGamePlies`, `LoadOptions.onLogicalLoaded`, and the permanent `browser_id` storage.

**Revert the id-space change too.** `isEngineGameIdTaken` and `genUniqueGameID`'s extra check
([gamesManager.ts:71](src/server/database/gamesManager.ts#L71)) go away, along with the new
`issueUniqueGameId` export from `gamemanager` — engine games get their id from the normal creation
path, and an `engine_games` row is keyed to a `games` id that already exists.

**And `createenginegame.ts`'s cooldown and daily cap.** They're bypassed by clearing one cookie and
cost a full table scan per creation. With creation going over the already-rate-limited socket, game
state changing only through validated move submissions, and `membersInActiveGames` capping
concurrency at one game per identity, there's nothing left for them to protect.

Creation still happens over the socket (`createengine`), but it now calls the normal game-creation
path. Loading is the game page's existing live path — `onlinegame.subscribeToGame()` → `gamestate`
→ `loadGameFromState` — plus spinning up the engine worker. No HTTP endpoints, no resume state, no
clock reconstruction: the server sends clock values like it does for every other live game.

### 1.2 Two new tables, mirroring the two that exist

`games` never tracked participants — `player_games` does, one row per participant, absent = guest.
`player_games` can't host an engine (it's welded to `user_id`, elo, rating deviation), so the engine
participant gets its own participant table, and a live counterpart:

```sql
engine_games (                          -- mirrors player_games
  game_id             INTEGER NOT NULL,
  player_number       INTEGER NOT NULL,
  score               REAL,             -- NULL for aborted, as in player_games
  clock_at_end_millis INTEGER,
  engine              TEXT NOT NULL,
  engine_version      TEXT NOT NULL,
  strength_level      INTEGER NOT NULL,
  PRIMARY KEY (game_id, player_number)
)

live_engine_games (                     -- mirrors live_player_games
  game_id           INTEGER NOT NULL,
  player_number     INTEGER NOT NULL,
  time_remaining_ms INTEGER,
  engine            TEXT NOT NULL,
  engine_version    TEXT NOT NULL,
  strength_level    INTEGER NOT NULL,
  PRIMARY KEY (game_id, player_number)
)
```

- **`score` is not redundant** even though a 2-player game's engine score is derivable from the
  human's — it stops being derivable for engine-vs-engine or 4-player games.
- **No disconnect columns on `live_engine_games`.** An engine never disconnects; that asymmetry is
  the reason this is a separate table rather than nullable columns on `live_player_games`. Say so in
  a comment.
- **`engine_version`** comes from `manifest['engineVersion']`, recorded at creation and carried to
  `engine_games` at conclusion like every other live→dead field. Right now that value is unreachable
  from game logic: `configureNunjucks` reads `dist/manifest.json` into a local and hands it straight
  to `setManifestGlobals`, so the only consumer is the template engine — and game creation must not
  import the view-engine config to get at it. **Move the manifest load into its own module**
  (`src/server/config/manifest.ts`): the parsed record in a module variable, `loadManifest()` keeping
  today's throw-if-absent behaviour, plus `getManifest()` and `getEngineVersion()`. `nunjucks.ts`
  becomes a consumer of it and points its dev `fs.watch` at the module's reload. That's a move, not
  an addition — no behaviour change and no new build output. While you're there, BUILD.md's manifest
  section describes only logical-name→hashed-path entries; note that it also carries `engineVersion`.
- **No "is engine game" flag on `live_games`/`games`.** Presence of a `live_engine_games` /
  `engine_games` row _is_ the flag; a boolean column would be a second source of truth that can
  disagree with the rows, and there's no case where it could be right while they're wrong.
  `liveGameRestore` bulk-loads its rows already — add one `getAllLiveEngineGames()` and group by
  `game_id` in memory, so startup stays one query per table.
- **In memory it does need a marker**, though — no flag in the database, one field on the game. The
  new code paths in `gamemanager` / `gameutility` / the disconnect logic have to know, so add to
  `MatchInfo`:

    ```ts
    /** Present only for games against an engine. Its moves arrive over the human's socket. */
    engineParticipant?: { color: Player; engine: ValidEngine; version: string; strengthLevel: number };
    ```

    Add a `gameutility.isEngineGame(servergame)` predicate for the branch sites, matching the existing
    `isGameOver` / `isGameResignable` style, and read the field directly where the data itself is
    needed. Put it on `MatchInfo` rather than inside `playerData`, because `PlayerData.identifier` is
    `AuthMemberInfo` and widening that union to admit an engine ripples through `activeplayers`,
    disconnect, logging and rating. It's reconstructed from the `live_engine_games` row on restore —
    which is exactly why the database needs no flag of its own.

Everything downstream then comes free: `gamelogger` writes `games`/`player_games`/`player_stats`
exactly as for PvP (stats are tracked like any other game), `restoreLiveGames()` handles server
restarts, and the dead-game readers need no engine special-casing beyond joining the new table for
the participant's name — so `deadgamestate`'s `assembleStaticGameState` keeps a small hook, but
`produceEngineGameStaticState` / `produceEngineGameResumeState` / `getLiveEngineGame` /
`isEngineGameOwner` all disappear.

### 1.3 Engine-specific behaviours to implement

| Situation                                  | Required behaviour                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
| ------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Owner disconnects on the **engine's** turn | Freeze the clock — the engine can't think while the tab is closed. Also `clearTimeout(servergame.match.autoTimeLossTimeoutID)`, which runs independently of the clock values. On reconnect, restore to `clocks.timeRemainAtTurnStart` (the search restarts from scratch, so the spent time is genuinely lost) and re-arm the forfeit timer — `liveGameRestore`'s `pendingTimers.autoTimeLossMs` already computes this "re-arm after a gap". **No new state is needed;** `ClockData` already carries `timeRemainAtTurnStart`.                                                                                                                                                                                                                                                                                                                                                       |
| Owner disconnects on **their own** turn    | Clock keeps ticking; they can forfeit on time, exactly as in PvP.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
| Owner gone for 5 minutes                   | Conclude — reusing `timeBeforeBothDisconnectedAutoConclusionMillis`. **Victor is the engine**, not a draw: PvP awards a draw by abandonment because neither player can be blamed, but here only the human can abandon, so a draw would let anyone escape a losing game. Abort instead if the game isn't yet resignable (< 2 plies).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
| Engine fails to produce a move             | The engine **loses**, which is what [enginegame.ts:236](src/client/scripts/esm/game/misc/enginegame.ts#L236) already concludes (`{ condition: 'resignation', victor: human }`) — a human whose client breaks abandons and loses rather than drawing, and it leaves the user with a finished game instead of a stuck board. What it needs is a channel: add one `engineresign` action to the existing `game` router union, handled by `resignGame(servergame, engineColor)`, which already awards the win to the opponent of the role it's given. Fall back to `abortGame` when the game isn't yet resignable (< 2 plies) — the same rule the human's own resign/abort buttons follow, so neither function's semantics change. No retry logic.                                                                                                                                      |
| Engine's move submission                   | Both colors' moves arrive over the owner's socket, so **derive** the role rather than adding a schema field: `role = engineParticipant ? servergame.whosTurn : ws.metadata.subscriptions.game.color`. That grants no new capability — `whosTurn` flips after every accepted move, so the owner still can't move twice for one side — and the subscription check still confines it to the game's human participant. Comment that the "is it their turn" check is vacuous for engine games; the move-number check still catches desync. **Also skip the opponent-move send in `broadcastMove`:** it targets `invertPlayer(role)`, which for an engine move is the human who just submitted it — echoing it back would double-apply and desync them. Spectators still receive it. `doesGameConclusionCheckOut`'s "can't submit a move where your opponent wins" rule works unchanged. |
| Cheat reports / rematch                    | Exempt engine games. Draw offers need nothing — the game page already omits them, so `onOfferDraw` never fires.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |

Accepted consequences, for the record: mid-game resumability becomes 5 minutes instead of 30 days;
a live engine game occupies the one-game-at-a-time slot, so **the lobby must also disable "Play
against computer"** while the user is in a game, alongside the two buttons it already disables; and
the engine is charged for network latency on its clock, as lichess bots are.

### 1.4 Also verify

`commitMove`'s local-clock branch (`!gamefile.untimed && gamesession.getGameType() === 'engine'` →
`clock.push`) exists because engine games had no server clocks. Once they do, engine games take the
online clock path and that branch — plus the `'moves-changed'` reorder this PR made to support it —
may be dead. Check before leaving it in. Board-editor engine games keep it alive until
[eactions.ts](src/client/scripts/esm/game/boardeditor/actions/eactions.ts) moves to the
handoff-to-lobby flow.

---

## 2. Bugs to fix regardless of architecture

### 2.1 The engine plays on after the game ends, un-ending it

Resign during the engine's turn and a few seconds later it moves anyway, restarting the clocks and
letting play continue. Three separate defects:

1. `enginegame.ts` contains **no `terminate()` call at all** — the worker outlives the game. Kill it
   on `'game-concluded'` and `'game-unloaded'`. `ceval.ts:297/343` and `gamereview.ts:499/535`
   already establish this pattern, and terminating is what actually stops the rayon threads.
2. `handleEngineMessage` / `makeEngineMove` never check `gamefile.gameConclusion`. **Reject any
   engine reply that arrives after the game is over, and `console.error` it** so future variants of
   this bug surface instead of silently resurrecting games.
3. `onMovePlayed` checks `gameConclusion` _after_ calling the human-move hook. Move the check to
   immediately before `postMessage`, so a conclusion landing between the move and the request can't
   slip through.

### 2.2 Custom-position engine games load with an out-of-range world border

Two implementations of the same feature, and they disagree. The existing one is
`Additional.worldBorderDist` → [boardpreviewer.ts:112-130](src/shared/chess/logic/boardpreviewer.ts#L112-L130)
("bounding box ± dist, only if `gameRules.worldBorder` is unset"). The PR adds a second,
[apeiron_card.setDefaultWorldBorder](src/shared/chess/engines/apeiron_card.ts#L203), which does the
same thing **plus clamps every edge to `±BORDER_CAP`**:

| Path                                                                                                           | Mechanism                                      | Capped? |
| -------------------------------------------------------------------------------------------------------------- | ---------------------------------------------- | ------- |
| Board editor ([eactions.ts](src/client/scripts/esm/game/boardeditor/actions/eactions.ts))                      | `setDefaultWorldBorder` writes an exact border | ✅      |
| Modal validation ([gameSetupModal.ts:211](src/client/scripts/esm/views/index/gameSetupModal.ts#L211))          | `setDefaultWorldBorder` on a copy              | ✅      |
| **The actual game load** ([enginegameloader.ts:83](src/client/scripts/esm/views/game/enginegameloader.ts#L83)) | `worldBorderDist` → boardpreviewer             | ❌      |

So the comment _"check it the same way the game will load it"_ is false, and
`isPositionSupported` — which rejects any edge beyond `BORDER_CAP`
([apeiron_card.ts:99-106](src/shared/chess/engines/apeiron_card.ts#L99-L106)) — validates the capped
copy while the game loads the uncapped one. Preset variants are safe because the `±2000n` cushion in
`engineDictionary` was chosen so every preset's pieces fit inside it, but paste a custom ICN with
pieces around x = 10 000 and the loaded border becomes `I64_MAX + 8000`, outside the engine's range,
with nothing left to catch it.

**Fix:** collapse to one implementation. Keep `worldBorderDist` and give `boardpreviewer` an
optional clamp (e.g. `Additional.worldBorderCap?: bigint` — "clamp each computed edge to ±cap"), then
delete `setDefaultWorldBorder` and the duplication in `eactions.ts`. Validation and loading then
agree by construction rather than by hand.

### 2.3 Worker init failure is silent

[apeiron.ts:75-77](src/client/scripts/esm/game/chess/engines/apeiron.ts#L75-L77) catches an init
failure, logs to the console, and posts nothing — so nothing on the page ever learns the engine is
unusable and the user gets no error.

The analysis page already handles this, in three places, all driven by the worker posting
`{ type: 'initerror', message }`: [ceval.ts:323](src/client/scripts/esm/views/analysis/ceval.ts#L323)
terminates the legal-moves worker, [ceval.ts:431](src/client/scripts/esm/views/analysis/ceval.ts#L431)
runs `handleWorkerFailure` (terminate both workers, `enabled = false`, `notifyStatus('failed')`), and
[gamereview.ts:511](src/client/scripts/esm/views/analysis/gamereview.ts#L511) runs
`handleWorkerFault`. So the game page having none is the asymmetry.

**Fix:** match them — post `initerror`, terminate the worker, and let the failure reach the page's
standard load-failure path. Nothing more than the analysis page does. It's fine that the load never
completes (the game is unplayable either way); it must not fail silently.

### 2.4 starfield: a cast-and-guard where the cause is call ordering

[starfield.ts:320-321](src/client/scripts/esm/game/rendering/starfield.ts#L320-L321) casts
`gboundingBox()` — declared `: BoundingBox` — to `BoundingBox | undefined` and guards. The cast
contradicts the signature, so it will rot from either side.

The cause is ordering: `starfield.update()` runs at
[gamecore.ts:147](src/client/scripts/esm/game/chess/gamecore.ts#L147), twenty lines _before_
`boardgeometry.recalcVariables()` at [gamecore.ts:167](src/client/scripts/esm/game/chess/gamecore.ts#L167),
so it reads last frame's box every frame and an unassigned one on frame 1. The file already
establishes the fix one line below the recalc, where `effectZoneManager.update()` is deliberately
placed _after_ it with that exact comment.

**Fix:** move `starfield.update()` below `boardgeometry.recalcVariables()`; delete the cast, the
guard, and the new `BoundingBox` import.

---

## 3. Redundancy with systems we already have

### 3.1 `EngineGameHooks` re-invents `GameBus`

[enginegame.ts:39-57](src/client/scripts/esm/game/misc/enginegame.ts#L39-L57) +
[registerHooks](src/client/scripts/esm/game/misc/enginegame.ts#L296) + the top-level registration in
[checkmatepractice.ts:75-82](src/client/scripts/esm/game/chess/checkmatepractice.ts#L75-L82) add a
bespoke single-slot observer registry so `enginegame` doesn't import the practice page's module. The
decoupling goal is right; the mechanism is redundant — `GameBus` is exactly this, and `enginegame`
is already listening to it two lines above the hook it forwards to. It's also a footgun:
`hooks = newHooks` means a second registrant silently erases the first.

**Fix:** delete the interface, the module variable, the setter and the export.
`onGameConclude` → `checkmatepractice` listens to `'game-concluded'` itself (it already re-checks
`inCheckmatePractice` internally). `onHumanMove` → `'user-move-played'`. `onEngineMove` → add one
`'engine-move-played'` event to `GameBus`, beside the existing `'opponent-move-played'`, and dispatch
it from `makeEngineMove`.

### 3.2 Don't add a third engine loader — and migrate the two that exist

`enginegameloader.ts` re-implements `gameloader.startEngineGame`'s whole sequence
(`setSessionGame` → `loadGamefile` → `concludeGameIfOver` → `await initEngineGame` →
`markLoadingDone` → `onCatchLoadingError`) as a third copy. Under §1 most of it goes away: the game
page's existing live path loads the game, and the only engine-specific work left is spinning up the
worker.

Separately, and per the redesign's per-page migration policy: `gameloader.ts` is the deprecated
`main`-branch loader, and its two engine functions should be **moved out and deleted from it**, not
imported by new code. `startEngineGame`'s only consumer is the checkmate-practice page, so it belongs
there; `startCustomEngineGame` disappears once
[eactions.ts](src/client/scripts/esm/game/boardeditor/actions/eactions.ts) adopts the
analysis page's "continue from here" handoff (stash the position, navigate to the lobby, auto-open
the modal — for the board editor, in "Play against computer" mode). Mirror how the game and analysis
pages own their loaders; don't reach back into `gameloader.ts`.

### 3.3 Duplicated wasm bring-up, and a second thread policy

- [apeiron.ts:59-78 `initWasm`](src/client/scripts/esm/game/chess/engines/apeiron.ts#L59-L78) is a
  line-for-line copy of
  [apeironanalysis.worker.ts `initialize`](src/client/scripts/esm/views/analysis/apeironanalysis.worker.ts#L139):
  computed-URL dynamic import → `wasm.default()` → `typeof wasm.initThreadPool === 'function'` guard
  → `initThreadPool`. The wasm-bindgen glue itself must stay unbundled, but this wrapper around it
  needn't be duplicated. Extract `loadEngineWasm(engineUrl, threads)` into
  **`src/client/scripts/esm/game/chess/engines/enginewasm.ts`** and import it from both workers (the
  analysis page already imports freely from `game/`). Both are ESM entry points with
  `splitting: true`, so it becomes a shared chunk — nothing to register in `ESMEntryPoints`, no
  duplication. Type the module's exports there with a small interface (`default`, `Engine`,
  `initThreadPool`) instead of `any`, so it's written once and both call sites keep type safety.
- [enginegame.getEngineThreadCount:290](src/client/scripts/esm/game/misc/enginegame.ts#L290) is a
  second thread policy alongside [ceval.ts:99-127](src/client/scripts/esm/views/analysis/ceval.ts#L99-L127),
  and they disagree: `min(4, max(1, hc - 1))` vs `clamp(hc, 1, THREAD_CAP = 4)`. ceval's feature
  detection is also stricter (`Atomics`, a shared `WebAssembly.Memory` probe). Export ceval's
  `BROWSER_SUPPORTS_THREADS` / `maxThreads()` and use them; if the `-1` is deliberate for gameplay,
  make that a documented parameter rather than a separate implementation. Either way an
  apeiron-specific Lazy-SMP detail doesn't belong in the engine-agnostic session module.

### 3.4 The preset variant-support check is unreachable

The preset branch of [isVariantSupportedByEngine](src/client/scripts/esm/views/index/gameSetupModal.ts#L213-L225)
can never fire: `setEngineOnlyVariants` hides every unsupported preset and coerces an unsupported
selection to Classical on every `openModal`, and the handoff path only pre-fills ICN.

**Fix:** delete the preset branch and its toast. `getSelectedVariantOptions()` then collapses to
"custom options, or null", losing the ambiguous double meaning that currently forces the caller to
re-query `getSeekVariant()` to tell "preset" from "invalid" apart. The custom-position check stays as
immediate UX feedback.

No server-side equivalent is needed. The only route to an unsupported engine game is a hand-crafted
message, it harms nobody but the sender, and the failure already resolves cleanly — the engine can't
produce a move, so it resigns and loses.

---

## 4. Dead code and unreachable guards

Each of these tells the next developer to plan for a scenario that can't occur:

| Where                                                                                                                                  | What                                                                                       | Why it can't happen                                                                                                                                                                                                                                                                                                     |
| -------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| [databaseTables.ts:667](src/server/database/databaseTables.ts#L667) + [:438](src/server/database/databaseTables.ts#L438)               | `addTurnStartTimeColumnToEngineGamesIfNeeded`                                              | A migration for a table created in the same PR; `generateTables()` already includes the column. No DB can lack it.                                                                                                                                                                                                      |
| [guigameactions.ts:229](src/client/scripts/esm/game/gui/guigameactions.ts#L229)                                                        | `if (gamefile.gameConclusion) return;`                                                     | `concludeGame()` dispatches `'game-concluded'` synchronously → `refresh()` → `showOnly(ActionsOver)`, hiding the resign/abort block. There's even an `armGracePeriod` for split-second clicks.                                                                                                                          |
| [enginegames.ts:420](src/server/game/gamemanager/enginegames.ts#L420)                                                                  | all of `resolveDeadEngineGameParticipantColor`                                             | It exists only to identify a dead **guest** owner from a permanently stored `browser_id`, which §5 drops — so delete the function and revert both `gamePageController` call sites to leaving guests unresolved. A signed-in owner still resolves through their `player_games` row, as in PvP.                           |
| [enginegamesync.ts:41](src/client/scripts/esm/views/game/enginegamesync.ts#L41)                                                        | `if (!gamefile) return;`                                                                   | `'moves-changed'` only fires with a loaded gamefile.                                                                                                                                                                                                                                                                    |
| [apeiron.ts:89-93](src/client/scripts/esm/game/chess/engines/apeiron.ts#L89-L93)                                                       | `if (!wasmInitialized)`                                                                    | The page only sends requests after `'readyok'`. It replaced a lazy-init fallback that did something; this only logs.                                                                                                                                                                                                    |
| [enginegame.ts:106](src/client/scripts/esm/game/misc/enginegame.ts#L106) + [:127](src/client/scripts/esm/game/misc/enginegame.ts#L127) | `options.workerUrl ?? '…/${currentEngine}.js'`, and `if (options.engineUrl !== undefined)` | The only other engine, `engineCheckmatePractice`, is commented out of `ESMEntryPoints` — it isn't built, so the "legacy" path 404s. Make both params required; a compile error is more honest than a silent 404.                                                                                                        |
| [types.ts:352](src/shared/types.ts#L352), [:355](src/shared/types.ts#L355), [:358](src/shared/types.ts#L358)                           | `ENGINE_GAME_MOVES_STRING_CAP`, `ENGINE_GAME_MAX_PLIES`, `countEngineGamePlies`            | Caps don't prevent DB flooding — a spammer creates many _small_ games. All they do is break the rare legitimate long game: past 2 000 plies the conclude POST 400s and the game is never saved, with one `console.error`. The real protection is the per-move distance-vs-duration rule already in `movesubmission.ts`. |

Also drop, with the endpoints they belong to: `IMPOSSIBLE_ENGINE_CONDITIONS`
([enginegames.ts:238](src/server/game/gamemanager/enginegames.ts#L238)) and the unparseable-moves
throw — both only reachable from a hand-crafted request, and both 500 rather than reject.

---

## 5. Smaller items

- **`browser_id` must not be persisted permanently.** `live_player_games.browser_id` exists but is
  transient (the row dies with the game); `engine_games` would be the first table to keep browser ids
  forever, purely to orient a dead guest's board — which we don't do for PvP guests either. Under §1
  this resolves itself: `addUserToActiveGames` takes `AuthMemberInfo`, where `browser_id` is
  guaranteed, so the `?? ''` fallbacks disappear by construction, and
  `resolveDeadEngineGameParticipantColor` goes with it (see §4).
- **COEP on `/game/:id` is fine, but its documentation is now wrong.** The
  `crossOriginIsolation` middleware's JSDoc says _"Applied ONLY to the analysis page: `require-corp`
  would block cross-origin subresources that don't send CORP (Turnstile, YouTube embeds, the
  analytics beacon) on other pages."_ Nothing on the game page loads a cross-origin subresource, so
  route-wide is the simpler choice — just rewrite the comment to say the game page is included and
  why that's safe.
- **`variantselector.css`'s `.hidden` override.** Every other stylesheet ends with its own section
  containing a plain `.hidden { display: none; }` and a comment about `display` rules in that file
  overriding `global.css` — source order does the work. Do the same here instead of the compound
  `.variant-group-item.hidden, .variant-item.hidden`; it's the house pattern and it also covers
  `.variant-list-back` and anything added later.
- **`setEngineOnlyVariants`** uses a document-wide `document.querySelector('button[data-group=…]')`
  when the module already keeps a group-keyed map two functions up; and the
  `if (group === 'custom') return` early-out sits _after_ the item loop it reads as though it skips.
- **`openModal`'s comment** says _"Computer games can only be preset variants the engine supports"_ —
  the code allows custom positions too.
- **The random color should be rolled server-side**, like the seek path does (`color: null` → server
  decides), rather than `Math.random()` in `handleComputerGame`. Under §1 creation is server-side
  anyway.
- **The worker file should be `apeiron.worker.ts`.** Every other worker uses that suffix
  (`apeironanalysis.worker.ts`, `icnvalidator.worker.ts`); the engine directory is the only holdout,
  and this PR is what turns `apeiron.ts` into a built worker entry point, so it's the moment to
  rename it. Note the knock-on: `game.njk` builds its manifest key by concatenating
  `'scripts/esm/game/chess/engines/' + engine + '.ts'`, which the rename breaks — another reason to
  resolve the URL in the controller (below) rather than assembling the key in the template.
- **`game.njk` patches `window.gamePageData` after the JSON dump**, and
  [globals.d.ts](src/client/types/globals.d.ts) declares two fields the server's `GamePageState`
  type doesn't — the same object is now typed in two places that disagree. `analysis.njk` avoids
  this by putting the manifest lookups inside the object literal; cleanest is resolving both URLs in
  the controller so there's one declaration.

Naviary2:
# Review — PR #1144 "Engine games on the game page"

## Verdict

The architectural rewrite landed and it is faithful: the parallel engine pipeline is gone, engine
games are ordinary `ServerGame`s in `activeGames`, the two new tables match the agreed schema column
for column, and every symbol that was supposed to disappear is gone from the tree. `type-check`,
`lint` and `npm test` (7 files, 83 tests) all pass. Nearly all of the review's individual items are
genuinely closed.

It is **not mergeable yet**. The largest problem is scope: the PR also builds full server-side
support for **custom-position live games** — for player-vs-player as well as player-vs-engine — which
was never asked for and needs its own design and its own review (§1). That has to come out. Beyond
it: a temporary DB migration was added for a table this same PR creates, which is the exact defect
the review asked to be deleted (§2); a stale clock-stamp write puts the human's remaining time on
every engine move (§3); the draw actions are ungated, so a forged `offerdraw` stores PvP-only state
in an engine game (§4); and spectators boot a full wasm engine and thread pool that never computes
anything (§6). The rest are cleanups.

Two commits added after that assessment have to come out wholesale before any of it is worth
addressing — see **Revert first** below.

## Scope

Do not add more than what the action items below ask for. Keep things in scope to the features this
PR has already added. Do not expand.

## Reporting back

When the work is done, post a written summary of it. Not a walkthrough of the diff — a short account,
enough that the changes don't have to be reconstructed by reading commits:

- Each item below: done, done differently, or not done. One line of reasoning for anything that isn't
  a plain "done".
- Confirmation that the two commits named under **Revert first** were reverted whole, not partially
  edited, and that nothing from either was carried forward.
- Anything changed that no item here asked for, and why it was necessary. They must have extreme justification.

---

## Revert first

Revert `3f400f76b` "Improve engine game startup" and `8fb186740` "replay local moves missing from the
server after reconnect with clocks" **in full**, before working through anything below.

**Keep: nothing from either commit.** Everything earlier in the branch stands as assessed in the rest
of this document. Revert only these two.

Every line reference in this document is against the tree **after** that revert. Several are off by a
dozen lines against the current head, because these two commits inserted into the same files.

### `8fb186740` — the client is never authoritative over the clocks

An engine game continuing locally while the socket is down is fine, and so is replaying those moves
once it reopens. What this commit adds on top is not: it reconstructs the server's clocks from the
client's own values on reconnect, through a new `syncclocks` action
([gamerouter.ts:36](src/server/game/gamemanager/gamerouter.ts#L36)) whose handler
([gamemanager.ts:414-442](src/server/game/gamemanager/gamemanager.ts#L414-L442)) overwrites
`clocks.currentTime` from the message. Its one sanity bound is
`startTime.millis + movesPlayed * increment` — which is "I have consumed zero time" — so a
hand-crafted client restores its clock to full at any moment, and the value flows on into
`armAutoTimeLoss`, the `live_engine_games` row and `clock_at_end_millis`. Engine games write
`player_stats` like any other game, which is exactly why client-asserted clock state is
disqualifying.

Tightening the bound is not the fix; the message should not exist. The clocks are the server's alone
and it advances them only when it itself receives a move submission, so whatever wall time elapsed
during the outage is charged to whichever colour was ticking when the socket dropped — derived
server-side, requiring nothing from the client, and never resolving in the offline player's favour.

Do not rebuild the replay path in this PR. It will be designed separately.

The rest of the commit exists to serve the `syncclocks` mechanism and goes with it:

- `clock.push` on the client for engine games
  ([movesendreceive.ts:46-49](src/client/scripts/esm/game/misc/onlinegame/movesendreceive.ts#L46-L49))
  — this is the local-clock branch §1.4 removed from `commitMove`, relocated into `sendMove` and no
  longer even gated on being offline. The client pushes clocks in **no** game.
- The second resync early-exit
  ([resyncer.ts:88-103](src/client/scripts/esm/game/misc/onlinegame/resyncer.ts#L88-L103)), which
  shadows the existing one-move resubmit path directly below it.
- The `socketman.getSocket()?.readyState` clause
  ([movesendreceive.ts:50](src/client/scripts/esm/game/misc/onlinegame/movesendreceive.ts#L50)):
  `socketmessages.send` already reopens a closed socket through `establishSocket`, and
  `SocketBus 'closed'` already sets `inSync = false`, so the clause is unreachable and suppresses the
  reconnect it appears to guard.

### `3f400f76b` — the engine gets no privilege the human doesn't have

It adds an `engineready` action so the engine's clock doesn't start until the client's wasm has
finished loading, deleting `resumeEngineClock` from
[onSubscribe.ts:28](src/server/game/gamemanager/onSubscribe.ts#L28) in exchange. Two reasons it goes:

- **The premise is wrong.** A human whose page takes three seconds to load is charged those three
  seconds. Nothing in the codebase lets a client withhold its own clock, and exempting the engine
  treats it better than the player it's playing. There is barely a gap to close in any case: the
  engine is bounded by `defaultTimeLimitPerMoveMillis` rather than by its clock, and freeze/resume is
  already lossless across refreshes — `freezeEngineClock` writes
  `currentTime[engine] = timeRemainAtTurnStart` and `resumeEngineClock` reads it back, so boot time is
  refunded on the next disconnect. Not worth permanent wire surface.
- **As written it breaks §1.3's reconnect behaviour.** `engineready` is sent once, from
  `initEngineGame`, reachable only via `loadGameFromState` — "a fresh page load (**not** a
  reconnect)". A socket blip on an open page therefore never re-sends it, `resumeEngineClock` never
  runs, the frozen clock never restarts and the forfeit timer is never re-armed. When the engine's
  move eventually arrives, `clock.push` skips its entire body because `timeAtTurnStart` is undefined
  ([clock.ts:142](src/shared/chess/logic/clock.ts#L142)), so the engine's think time is never
  deducted. Repeatable: drop the connection each time the engine starts thinking and it plays the
  whole game on its opening clock.

`resumeEngineClock` itself is correct and goes back to being called from `onSubscribe` — with §7's
`isGameOver` guard added to it.

Also in this commit and also out: `computer_engine_strength` as a persisted preference (17 lines in
[preferences.ts](src/client/scripts/esm/components/header/preferences.ts), plus
`setSelectedEngineStrength` and a `data-level` special case inside the generic toggle-group loop in
[gameSetupModal.ts:96](src/client/scripts/esm/views/index/gameSetupModal.ts#L96)). Persisting the
modal's choices is planned as one change covering **every** option in it, not one row at a time — a
strength-only version now is a half-feature the general one would have to undo, and it sets the
precedent of each row growing its own bespoke persistence.

Its default also moves the modal's pre-selected strength from 4 to 8, putting a new player's first
computer game against the strongest engine — a separate product decision, and not this PR's to make.
[index.njk:240](src/server/views/index.njk#L240) goes back to `active` on level 4.

One pre-existing comment is worth correcting while here, because it is what produced the unit bug
above: [clock.ts:36](src/shared/chess/logic/clock.ts#L36) documents
`ClockData.startTime.increment` as "in milliseconds". It is in **seconds** —
[clockutil.ts:88](src/shared/chess/util/clockutil.ts#L88) takes it straight from the `s+s` time
control's second field, and every correct reader wraps it in `timeutil.secondsToMillis`.

---

## Not fixed

### 1. Out of scope: the PR implements custom-position live games, for PvP as well as engine games

On `redesign`, starting a game from a custom position is deliberately unimplemented. Both entry
points throw the same sentence:

- [gamemanager.ts:85](src/server/game/gamemanager/gamemanager.ts#L85) — `'Custom variant game starting is not yet implemented.'`
- [gameutility.ts:268](src/server/game/gamemanager/gameutility.ts#L268) — same

A custom seek can be _posted_ and travels over the wire intact, but accepting one throws. This PR
removes both throws and builds the missing pipeline:

| File                                                 | What was added                                                                                                                                                                                                                                                                       |
| ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `gameutility.ts`                                     | `MatchInfo.variant: VariantCode \| null` + `position?: string`; `initMatch` custom branch; `initServerGame` takes `LoadedVariant \| undefined`; `buildStaticGameSetup` emits `{kind:'custom'}`; `getRatingDataForGamePlayers` null → INFINITY; `buildMetadataOfGame` omits `Variant` |
| `gamemanager.ts`                                     | `createGame` parses the ICN into `VariantOptions` and threads it into `initGame`                                                                                                                                                                                                     |
| `liveGameValues.ts`                                  | persists `position: match.position ?? null`                                                                                                                                                                                                                                          |
| `liveGameRestore.ts`                                 | reconstructs a custom game from `live_games.position`                                                                                                                                                                                                                                |
| `gamelogger.ts`                                      | `getICNOfGame` re-parses the position into the archived ICN; null-variant guards on both leaderboard writes                                                                                                                                                                          |
| `ratingabuse.ts`, `statlogger.ts`                    | null-variant guards (`?? 'Custom'`)                                                                                                                                                                                                                                                  |
| `onRematch.ts`                                       | rematch of a custom game                                                                                                                                                                                                                                                             |
| `onlinegame.ts`, `types.ts`, `gamePageController.ts` | client parses `gamePageData.position` into `variantOptions`                                                                                                                                                                                                                          |

Roughly 100 lines across ten files, and **none of it is gated on `engineParticipant`**. It sits in
the shared creation path, so [acceptseek.ts:106](src/server/game/seeksmanager/acceptseek.ts#L106)
handing a `{ kind: 'custom' }` seek to `createGame` now starts a real custom PvP game. Custom
player-vs-player games ship as a silent side effect of a PR about engine games.

Two arguments that might look like justification, and why neither holds:

- **"The board editor needs it."** The board editor lives in
  [play.ejs](src/client/views/play.ejs), which nothing renders — there are zero `.ejs` render calls
  left in `src/server`, and `/editor` serves [editor.njk](src/server/views/editor.njk), an 8-line
  stub with an empty `<main></main>`. Its "Start engine game" button reaches no user today.
- **"The review assumed it worked."** The review's §1 did say custom positions would "fall back to
  `applyClientReportedMove` like any PvP custom game". That assumption was wrong, but the correct
  response is to match PvP's _actual_ level of support, not to raise it.

**Action:** restore both throws and revert the custom-position plumbing listed above, including the
client half — `loadGameFromState`'s `variantOptions` parse, `GamePageData.position`, and
`gamePageController`'s `position: game?.match.position`. `liveGameValues.onGameCreated` goes back to
`position: null` with its original comment. Engine games against **preset** variants are unaffected
and work fully.

**Keep** exactly what PvP already has: `CreateEngineGameBodySchema.variant` stays `SeekVariantSchema`
so the lobby still sends the custom ICN over the wire, and `resolveAndValidateVariant` still resolves
cloud saves and validates position legality. A custom computer game then fails at `createGame`,
caught by `createEngineGameWs`'s existing try/catch and surfaced as an error toast — PvP parity, with
faster feedback than PvP's fail-on-accept.

**Leave the lobby modal alone.** Custom positions stay selectable in computer mode, exactly as custom
seeks stay postable in online mode. Do not extend `setEngineOnlyVariants` to hide the custom group.

**Do not revert** the engine-driven changes that happen to touch the same functions:
`buildMetadataOfGame`'s `getPlayerName` helper and its `white?.` / `black?.` optionality,
`initMatch`'s `engineParticipant` assignment, and `liveGameRestore`'s engine-row handling all stay.

**This also removes what would otherwise be a blocker.** With custom live games unreachable, no
`games` row can carry `variant = NULL`, so the dead-game load path can't reach its `{ kind: 'custom' }`
branch. As written, that branch crashes:
[onlinegame.ts:91-93](src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts#L91-L93) reads the
position from `window.gamePageData.position`, which the controller only populates for **live** games,
while the dead path ([deadgameloader.ts:42-52](src/client/scripts/esm/game/misc/onlinegame/deadgameloader.ts#L42-L52))
parses `deadState.icn` for the move list only and never forwards the position. Every concluded
custom-position game would have failed to load with "Failed to load game. Please refresh." Reverting
§1 makes it moot; when custom games are built for real, this is the trap to avoid — and
[deadgamestate.ts:171](src/server/game/gamemanager/deadgamestate.ts#L171) already documents the
intended contract ("its position comes from the ICN, parsed client-side"), which no client code
implements.

### 2. A temporary migration for a table this PR creates

[databaseTables.ts:455](src/server/database/databaseTables.ts#L455)
`migrateLegacyEngineGamesTableIfNeeded()` renames `engine_games` aside, rebuilds it, and back-fills
scores by joining `games.result` — about 35 lines of SQL in a transaction.

`engine_games` does not exist on `origin/redesign`. The only database that ever held the old
15-column shape is a developer's own working copy, which is deleted and regenerated, not migrated. On
production `db.columnExists('engine_games', 'game_id')` is false on first boot and the function is
unreachable forever.

This is the same defect the review listed as `addTurnStartTimeColumnToEngineGamesIfNeeded` ("A
migration for a table created in the same PR… No DB can lack it"), made substantially larger, and it
permanently tells the next developer that a pre-live-pipeline `engine_games` shape once shipped.

**Action:** delete the function and its call at
[databaseTables.ts:435](src/server/database/databaseTables.ts#L435).

### 3. Delete the client-side clock stamping

[onlinegamerouter.ts:208-213](src/client/scripts/esm/game/misc/onlinegame/onlinegamerouter.ts#L208-L213)
stamps the last move with `clockValues.clocks[ourColor]`, where `ourColor` is the viewer's own role.
That held while `'clock'` only ever arrived right after _we_ moved. In an engine game the owner
submits **both** colors, so a `'clock'` also arrives after the engine's move
([movesubmission.ts:152-157](src/server/game/gamemanager/movesubmission.ts#L152-L157)) and the
engine's move gets the human's remaining time written onto it.

**Action:** delete the three lines, keeping the `applyClockValues` call above them. Reasoning:

- The only reader of `move.clockStamp` for display is `guiclock.showViewedMoveClockStamps`, whose
  only caller is [guianalysisview.ts:33](src/client/scripts/esm/views/analysis/gui/guianalysisview.ts#L33)
  — the analysis page. `guiclock.update()` even early-returns on `getGameType() === 'analysis'`; the
  two paths are disjoint by design. **The game page never displays per-move clock stamps.**
- The analysis page loads from the archived ICN, where the server's own `moveRecord.clockStamp` is
  complete and correct. Nothing is lost.
- The only other reader is [copygame.ts](src/client/scripts/esm/game/chess/copygame.ts) via
  `gamecompressor`, and it has **zero callers** — the copy-game button isn't wired on the redesign
  game page yet.
- Inbound stamps stay: `'move'` messages, resync and `gamestate` all carry server-authored stamps for
  both colors, and those writes are correct.

Do **not** reach for `moveutil.getColorThatPlayedMoveIndex` here. The hazard isn't _which color_,
it's _which move_: the `'clock'` reply for ply N can land after ply N+1 is already applied locally,
so the correct number lands on the wrong ply. A correct fix needs a move index on the clock message —
schema complexity that buys nothing, given no consumer exists.

This is a **pre-existing latent bug**, not one this PR authored; engine games are simply the first
thing that submits two moves per round trip. It belongs in this PR because this PR is what makes it
fire.

### 4. The draw actions are ungated: a forged `offerdraw` stores PvP-only state in an engine game

`offerdraw`, `acceptdraw` and `declinedraw` are dispatched unconditionally at
[gamerouter.ts:95-103](src/server/game/gamemanager/gamerouter.ts#L95-L103), and no handler in
[onOfferDraw.ts](src/server/game/gamemanager/onOfferDraw.ts) checks `engineParticipant` — unlike
`report`, `claimvictory`/`claimdraw`, `offerrematch` and `subscriberematch`, which all refuse engine
games. `game.njk` omitting the draw button is client-side only, and a forged socket message bypasses
it.

A forged `offerdraw` at 2+ plies therefore reaches `openDrawOffer`, leaving an open draw offer on the
in-memory match and a `last_draw_offer_ply` write on the `live_games` row.

**Action:** add `if (gameutility.isEngineGame(servergame)) return;` as the first line of `offerDraw`,
`acceptDraw` and `declineDraw`, matching the four guards above.

---

## Fixed but introduced a new problem

### 5. Dead engine resign/abort path on the game page

[guigameactions.ts:203-232](src/client/scripts/esm/game/gui/guigameactions.ts#L203-L232) adds:

```ts
if (gamesession.getGameType() === 'engine') {
    // Engine games conclude locally; the sync module reports it to the server.
    ...
    concludeEngineGameLocally({ victor: engineColor, condition: 'resignation' });
    return;
}
```

Neither branch can run, and the comment describes an architecture that no longer exists:

- On `/game/:id` an engine game is session type **`'online'`**
  ([onlinegame.ts:86](src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts#L86)), so the
  condition is never true; resign and abort correctly go over the socket.
- The only remaining `type: 'engine'` session is checkmate practice
  ([checkmatepractice.ts:138](src/client/scripts/esm/game/chess/checkmatepractice.ts#L138)), and
  `guigameactions` is loaded solely by the game page
  ([game.ts:14](src/client/scripts/esm/views/game/game.ts#L14)) — `checkmatepractice.njk` renders a
  bare `<main>` with no action bar.
- "the sync module reports it to the server" refers to `enginegamesync.ts`, deleted in this PR.

There is also no behavioural gap it could have been covering.
[movesequence.ts:60-66](src/client/scripts/esm/game/chess/movesequence.ts#L60-L66) gates local
conclusion on `getGameType() !== 'online' && !== 'analysis'` — that is _all_ online games, validated
or not. Non-validated games don't self-conclude either; the conclusion always arrives from the
server, riding the opponent's `'move'`
([movesendreceive.ts:114-115](src/client/scripts/esm/game/misc/onlinegame/movesendreceive.ts#L114-L115),
comment: _"the server is boss, so if they say the game is over"_) or as a dedicated
`'gameconclusion'`. Engine games already match PvP exactly, and for them the owner receives the
conclusion via `broadcastMove`'s direct send to the submitter, since the opponent-`'move'` send is
skipped.

**Action:** delete both branches, `concludeEngineGameLocally`, and the `GameConclusion` / `typeutil` /
`gamesession` imports they pulled in. Runtime behaviour is unchanged.

### 6. Spectators boot a full wasm engine that never computes anything

[enginegame.ts:65](src/client/scripts/esm/game/misc/enginegame.ts#L65) made `youAreColor` optional —
_"Omit for spectators, who only use this worker for the legal-moves overlay"_ — and
[onlinegame.ts:124-143](src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts#L124-L143) passes
`youAreColor: ourRole`, which is `undefined` for spectators, with no gate. What a spectator of a live
engine game actually gets:

- The engine glue and `.wasm` downloaded, `wasm.default()` run, and the Lazy SMP pool brought up at
  `min(4, cores − 1)` threads.
- **No best-move request, ever.** `onMovePlayed` only fires on `'user-move-played'`, which spectators
  never dispatch; and even if it did, `engineColor` is `undefined` so `whosTurn !== engineColor`
  returns immediately.
- The one thing the worker can serve is `enginelegalmovesdebug`, which is `enabled = false` by
  default behind a debug toggle — the `requestMovesForCurrentPosition()` call at the end of init
  early-returns on `!enabled`
  ([enginelegalmovesdebug.ts:108](src/client/scripts/esm/game/misc/enginelegalmovesdebug.ts#L108)).

A full wasm and thread-pool boot that sits idle unless the viewer presses a debug key. Note that even
if it were used, ceval's legal-moves worker is inited with **no `threads` field at all**
([ceval.ts:318-322](src/client/scripts/esm/views/analysis/ceval.ts#L318-L322)) and so never brings up
a pool — legal-move generation needs no threads.

**Action:** don't initialize the engine for spectators. Make `youAreColor` required again and guard
the call site on `ourRole !== undefined`. This is a net deletion: the optional parameter and its
JSDoc, the `engineColor = options.youAreColor === undefined ? … : …` ternary, and the
`if (options.youAreColor !== undefined) onMovePlayed()` conditional all go away.

Lazy per-spectator init was considered and rejected: it would need a request queue, an in-flight
guard and a release path in `enginegame.ts`, mirroring ceval's `flushQueuedLegalMovesRequests` /
`terminateLegalWorker` machinery — a lot of spectator-specific state for a developer debug overlay.

### 7. `resumeEngineClock` restarts the clock on a concluded game

[gamemanager.ts:387-395](src/server/game/gamemanager/gamemanager.ts#L387-L395) guards on `untimed`,
`whosTurn` and `isGameResignable` — but not `isGameOver`. A concluded engine game lingers in memory
for the eviction window, and a page refresh during it sends a full `subscribe`
([onSubscribe.ts:28](src/server/game/gamemanager/onSubscribe.ts#L28)) — `game.ts:32` calls
`subscribeToGame()` before any state loads, so `stage` isn't yet `'finalized'` and the rematch path
isn't taken. `resumeEngineClock` then `clock.edit`s a finished game's clocks, broadcasts a bogus
`'clock'` to spectators, and issues UPDATEs against the `live_games` / `live_engine_games` rows that
`onGameLogged` already deleted. `armAutoTimeLoss` is the only thing that notices, because it checks
`isGameOver` itself.

**Action:** add `gameutility.isGameOver(servergame)` to the early-return.

The `colorTicking === engine.color` branch of the `remaining` ternary is **correct** and should be
left alone: it fires when a second tab takes over the game (`subscribeClientToGame` kicks the old
socket via `detatchSocketFromGame`, which doesn't run the freeze path). The new tab starts a fresh
worker whose search begins from zero, so restoring `timeRemainAtTurnStart` is right — the same
rationale as the disconnect case.

### 8. `logConcludedGame` hand-rolls a helper that already exists

[gamemanager.ts:542](src/server/game/gamemanager/gamemanager.ts#L542) replaced
`gameutility.broadcastToParticipants(...)` with an inline loop over `match.playerData`. But
[gameutility.ts:866-871](src/server/game/gamemanager/gameutility.ts#L866-L871) already skips players
with no socket and already iterates exactly that map — the engine case needed no change. The only
real difference is that the message moved from the `'game'` sub to `'general'`, which is where
`notifyerror` belongs everywhere else, so that part is a genuine fix worth keeping.

**Action:** give `broadcastToParticipants` a `sub` parameter and have it delegate to the existing
per-color sender, then call it with `'general'`:

```ts
// gameutility.ts
function broadcastToParticipants(
	servergame: ServerGame,
	sub: string,
	action: string,
	value?: any,
): void {
	for (const color of Object.keys(servergame.match.playerData))
		sendMessageToColor(servergame.match, Number(color) as Player, sub, action, value);
}

// gamemanager.ts
gameutility.broadcastToParticipants(servergame, 'general', 'notifyerror', message);
```

`sendMessageToColor` already handles the missing socket and the `'general'`/notify translation, so
the duplicated loop disappears rather than moving. Update the existing callers (including
`broadcastToAll`) to pass `'game'`.

### 9. Duplicated engine lookup on the dead-game page

For a concluded engine game, `getGamePageState` queries `engine_games` twice: once inside
`produceStaticGameState` → `produceDeadStaticGameState` →
[getEngineParticipant](src/server/game/gamemanager/deadgamestate.ts#L183), and again at
[gamePageController.ts:96](src/server/controllers/gamePageController.ts#L96) via
[getDeadEngineGameInfo](src/server/game/gamemanager/deadgamestate.ts#L115). The two readers are
near-identical row-0 mappers in the same file.

**Action:** delete `getDeadEngineGameInfo` and have `produceDeadStaticGameState` return the engine
info alongside `state` and `ratingChanges`, so the row is read once.

### 10. `Omit<>` — forbidden by `CLAUDE.md`

[liveEngineGamesManager.ts:19](src/server/database/liveEngineGamesManager.ts#L19):

```ts
type LiveEngineGameData = Omit<LiveEngineGamesRecord, 'game_id' | 'player_number'>;
```

**Action:** invert it the way the table this one mirrors does —
[livePlayerGamesManager.ts:16-27](src/server/database/livePlayerGamesManager.ts#L16-L27) declares
`LivePlayerData` and has `LivePlayerGamesRecord extends` it.

### 11. Four new files are missing their line-1 path comment

`src/server/config/manifest.ts`, `src/client/scripts/esm/game/chess/engines/enginewasm.ts`,
`src/server/database/engineGamesManager.ts` and `src/server/database/liveEngineGamesManager.ts` all
open straight into imports or a JSDoc block. `createenginegame.ts` has its header, so the convention
is followed inconsistently within one PR.

**Action:** add the path comment to all four, and give `enginewasm.ts` the short module-purpose
comment it currently lacks entirely.

### 12. An apeiron-specific branch is still in the engine-agnostic session module

The review's objection to `getEngineThreadCount` living in `enginegame.ts` was that "an
apeiron-specific Lazy-SMP detail doesn't belong in the engine-agnostic session module." The thread
_policy_ moved to `enginewasm.ts` correctly, but
[enginegame.ts:118](src/client/scripts/esm/game/misc/enginegame.ts#L118) now hardcodes
`if (options.currentEngine === 'apeiron')` to decide whether to post the init message — the same
coupling in a new form.

**Action:** add a boolean to the `Engine` interface in
[engine.ts](src/shared/chess/engine.ts) (e.g. `loadsWasmGlueAtRuntime`) and branch on
`engineDictionary[options.currentEngine]`. Separately, export `THREAD_CAP` from `enginewasm.ts`
alongside `maxEngineThreads` and have both `ceval.ts` and `enginegame.ts:305` use it instead of each
holding its own `4`.

### 13. Unrelated behaviour changes bundled in

Three changes have nothing to do with engine games. Verdicts:

- **Ping clamp — keep.** `Math.max(0, …)` in
  [onlinegamerouter.ts:174-178](src/client/scripts/esm/game/misc/onlinegame/onlinegamerouter.ts#L174-L178)
  is a genuine bug fix; a ping-adjusted clock should never go negative.
- **Ping rounding — delete.** Both `Math.round` calls in the same block. Clock values don't need
  rounding at this layer.
- **`toClockMillis` — keep.** It looked redundant with the ping rounding in place; with the rounding
  removed it becomes the only thing guaranteeing an integer reaches the wasm boundary
  (`wasm.Engine.from_icn(icn, engineConfig)` crosses into Rust). Its `Math.max(0, …)` is load-bearing
  either way, since [clock.ts:205-211](src/shared/chess/logic/clock.ts#L205-L211) lets `currentTime`
  go negative between the flag and the server's conclusion. Its `Number.isFinite` branch guards a
  case the types exclude and can go.
- **The `notifyerror` sub change — keep**, implemented per §8.

### 14. Time-control validation re-implemented instead of declared

[createenginegame.ts:29](src/server/game/seeksmanager/createenginegame.ts#L29)
`if (!clockutil.isTimedControlValid(body.timeControl)) return;` — a silent return that leaves the
client hanging with no feedback. `createseekschem` expresses the identical rule declaratively as a
zod `.refine` ([createseek.ts:66](src/server/game/seeksmanager/createseek.ts#L66)), so the router
rejects it uniformly.

**Action:** move it onto `CreateEngineGameBodySchema` as a `.refine` and delete the imperative check.

### 15. Smaller items

- **Drop `engine` from the creation schema — the server already knows it.** The client hardcodes
  `const COMPUTER_GAME_ENGINE: ValidEngine = 'apeiron'`
  ([gameSetupModal.ts:38](src/client/scripts/esm/views/index/gameSetupModal.ts#L38)) and sends it
  back, so a constant makes a round trip and returns trusted. `CreateEngineGameBodySchema.engine`
  ([types.ts:360](src/shared/types.ts#L360)) is `z.literal(Object.keys(engineDictionary))`, which also
  admits `'engineCheckmatePractice'` — whose `maxStrengthLevel: 1` passes the check at
  [createenginegame.ts:32](src/server/game/seeksmanager/createenginegame.ts#L32) — so a forged message
  creates a real live game and a permanent `engine_games` row for an engine that can't play one.
  **Action:** delete the field from the schema and the client body, and have `createEngineGameWs`
  supply the online engine itself, next to its existing `version: getEngineVersion()`
  ([createenginegame.ts:50](src/server/game/seeksmanager/createenginegame.ts#L50)). Name and version
  then come from the same authority. Move the constant into
  [engine.ts](src/shared/chess/engine.ts) rather than duplicating it server-side — the client still
  needs it for its own world-border validation
  ([gameSetupModal.ts:219](src/client/scripts/esm/views/index/gameSetupModal.ts#L219)), it just
  stops sending it. Add the field back as a real choice if a second online engine ever ships.
- **Toast when a failed engine load resigns the engine.** All three load failures —
  `!window.Worker`, `initerror` and the worker's `onerror` — resign the engine and reject
  `initEngineGame`, which is swallowed by
  [onlinegame.ts:140-142](src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts#L140-L142)'s
  `console.error`. The player gets a win they didn't earn with nothing on screen to explain it. It's
  rare (the assets are `immutable`-cached after the first engine game), so resigning is a fine
  resolution — it just needs to say so. **Action:** show an error toast in that one `.catch`, which
  every load failure already funnels through, and delete the now-redundant `alert()` at
  [enginegame.ts:87](src/client/scripts/esm/game/misc/enginegame.ts#L87).
- **Delete the orphaned translation keys.** `start_engine_game`, `start_engine_game_message`,
  `use_default_border`, `engine_difficulty`, `play_as` and friends
  (`translation/en-US.toml` lines 184-192) plus `position_not_supported` (line 540) are unreferenced
  now that `guistartenginegame.ts` and its `play.ejs` block are gone. **Action:** delete them from
  `en-US.toml` only — per `CLAUDE.md`, other languages aren't maintained by hand. Read
  [docs/systems/TRANSLATIONS.md](docs/systems/TRANSLATIONS.md) first.
- **Localize the engine-support toast.**
  [gameSetupModal.ts:224](src/client/scripts/esm/views/index/gameSetupModal.ts#L224) hardcodes
  `` `The engine doesn't support this position. ${result.reason}` ``; the flow it replaced used a
  translated string. **Action:** add a script-translation key and use it — the same file already
  localizes its submit labels via `t.index.lobby_buttons`. Read
  [docs/systems/TRANSLATIONS.md](docs/systems/TRANSLATIONS.md) first.
- **Delete all three redundant indexes**, including the pre-existing one. Verified by query plan: a
  composite `PRIMARY KEY (game_id, player_number)` on a rowid table creates `sqlite_autoindex_…_1`,
  and `WHERE game_id = ?` already resolves to `SEARCH … USING INDEX sqlite_autoindex_…`. The explicit
  index adds a second B-tree to maintain on every insert and buys nothing. This applies to
  `idx_engine_games_game`, `idx_live_engine_games_game` **and** the pre-existing
  `idx_live_player_games_game` (same PK shape). No migration is needed — an already-created index in
  production is harmless if simply no longer declared. `idx_player_games_game` **stays**: that table's
  PK is `(user_id, game_id)`, so without it SQLite genuinely does `SCAN player_games`.
- **Fix the stale JSDoc** on `onBothPlayersDisconnected`
  ([gamemanager.ts:595-598](src/server/game/gamemanager/gamemanager.ts#L595-L598)) — it still says
  "Concludes the game as a draw by abandonment"; engine games now award the win to the engine.
- **Finish the `commitMove` comment.**
  [movesequence.ts:53-55](src/client/scripts/esm/game/chess/movesequence.ts#L53-L55) explains why
  `'moves-changed'` must stay _above_ the game-over checks but not why it moved _below_ the
  `clockStamp` write — which is the load-bearing part, since
  [guiclock.ts:48/129](src/client/scripts/esm/game/gui/guiclock.ts#L48) reads `moves[i].clockStamp`
  on that event. The reorder itself is correct and also fixes an off-by-one in the PvP clock-stamp
  display; say so.

FirePlank:
1. Done — removed the out-of-scope custom-position live-game pipeline so custom engine seeks now fail at normal game creation, matching PvP support.
2. Done — deleted the temporary `engine_games` migration.
3. Done — removed client-side clock stamping.
4. Done — draw offer, accept, and decline ignore engine games.
5. Done — removed the unreachable local engine resign/abort path.
6. Done — spectators do not initialize the engine worker/wasm.
7. Done — `resumeEngineClock` now returns for concluded games.
8. Done — `broadcastToParticipants` accepts a route/sub and `notifyerror` uses the existing helper.
9. Done — dead-game engine data is fetched once and returned with static state.
10. Done — replaced the `Omit<>` type with interface inheritance.
11. Done — added the required path headers and enginewasm module comment.
12. Done — wasm-loading capability lives on `Engine` so `THREAD_CAP` is shared.
13. Done — retained the ping clamp and `toClockMillis`, removed ping rounding, and kept the `notifyerror` route correction.
14. Done — moved engine-game time-control validation to the creation schema.
15. Done — server-owned online engine selection, failed-load toast, translation cleanup/localization, redundant-index removal, and stale comments are addressed.

Reverts:

- `8fb186740` was removed in full so none of its client-clock or reconnect replay mechanism remains.
- `3f400f76b` was done differently: its strength preference/default-level changes were removed, while the engine-readiness clock gate remains. Its reconnect flaw was fixed by re-sending readiness after a reconnect and broadcasting the authoritative resumed clock to the player, so the clock visibly starts before the engine moves.

No unrelated behavior was retained. The only follow-up beyond the review items was making the retained readiness gate reconnect-safe and visually update the owner.

Naviary2:
# Review — PR #1144 "Add local engine (vs computer) games, saved like PvP games"

Reviewed at `53a5a1a78`, against base `redesign`.

`npm run type-check`, `npm run lint` and `npm test` (7 files, 83 tests) all pass.

The custom-position live-game pipeline is fully out — `MatchInfo.variant` is a plain `VariantCode`
again, both "not yet implemented" throws are back, `gamePageData.position` is gone, and
`gamelogger`, `statlogger`, `ratingabuse`, `liveGameRestore`, `onRematch` and `liveGameValues` are
byte-identical to `redesign` on those lines. The temporary `engine_games` migration is gone, the
draw actions are gated, spectators no longer boot a wasm worker, `resumeEngineClock` guards
`isGameOver`, the `Omit<>` is inverted, the redundant indexes are dropped, and the translation work
(new `[script.engine]` key, `interpolate`, regenerated types, English-only) is correct.

Six items remain. §1 is the blocker.

---

## 1. The `engineready` clock gate has to come out — the server is always boss of the clocks

**Files:** `enginegame.ts`, `resyncer.ts`, `gamerouter.ts`, `gamemanager.ts`, `onSubscribe.ts`

The engine's server-side clock currently does not restart until the client sends an `engineready`
socket message announcing that its wasm worker has finished loading. This needs to be removed
entirely, not made reconnect-safe. Three reasons, in order of weight:

**It breaks the contract that the server owns the clocks.** `resumeEngineClock` has exactly one
caller, [gamerouter.ts:93](src/server/game/gamemanager/gamerouter.ts#L93) — a client message. A
client that subscribes to a frozen engine game and simply never sends `engineready` leaves the game
with **no clock ticking for either side**: `armAutoTimeLoss` is never re-armed, no one can flag, and
the game plus its `live_games` row sit in memory until the process restarts. Every other clock
transition in the codebase is driven by something the server itself observed.

**It breaks PvP/PvE symmetry for no real gain.** A human whose page takes three seconds to load is
charged those three seconds; nothing lets a client withhold its own clock. And the gain is smaller
than it looks: a resumed engine does not resume its *search* — it restarts from depth zero and
remembers nothing it computed before the refresh. That loss was already accepted as the cost of
refreshing mid-turn. Refunding only the wasm-boot slice of it (and those assets are `immutable`-cached
after the first engine game) does not make the situation fair, it just adds a second special case.

**The complexity is not small.** Two module flags, a socket listener, an exported notifier and a
resync hook, versus one line at the subscribe site.

**Do this:**

1. Delete from [enginegame.ts](src/client/scripts/esm/game/misc/enginegame.ts): the `SocketBus`
   import (line 22), `engineReady` / `serverNeedsReady` (lines 48-49), the `'closed'` listener
   (lines 56-58), their assignments in `initEngineGame` (lines 90-91) and `terminate` (lines
   333-334), the `engineReady = true; notifyServerEngineReady();` pair (lines 136-137),
   `notifyServerEngineReady` itself (lines 145-150) and its default export (line 344).
   `onMovePlayed()` stays where it is.
2. Delete the `enginegame` import and the `enginegame.notifyServerEngineReady()` call from
   [resyncer.ts:26,62](src/client/scripts/esm/game/misc/onlinegame/resyncer.ts#L62).
3. Delete the `engineready` schema member ([gamerouter.ts:35](src/server/game/gamemanager/gamerouter.ts#L35))
   and its `case` (lines 92-94). The `case` also double-guarded `isEngineGame`, which
   `resumeEngineClock` already checks itself.
4. Restore the call in [onSubscribe.ts](src/server/game/gamemanager/onSubscribe.ts), on the
   participant path, immediately after `gameutility.subscribeClientToGame(game, ws, ourRole)` on
   line 27 — before the `gamestate` is built, so the reply already carries the resumed clocks.
   Import `resumeEngineClock` from `./gamemanager.js` there and drop it from `gamerouter.ts`'s
   import on line 22.
5. In `resumeEngineClock`, delete
   [gamemanager.ts:404](src/server/game/gamemanager/gamemanager.ts#L404) —
   `broadcastToParticipants(servergame, 'game', 'clock', clockValues)`. With the resume happening
   before `getGameStateMessageContents`, the reconnecting human already receives these values inside
   the `gamestate`; this line just duplicates them a moment later. The `broadcastToSpectators` call
   on line 405 stays: spectators are already attached and get no `gamestate`.

The `isGameOver` guard added to `resumeEngineClock` stays, and so does everything else in it.

---

## 2. The engine-load failure toast fires when the player ends the game themselves

**File:** [onlinegame.ts:132-137](src/client/scripts/esm/game/misc/onlinegame/onlinegame.ts#L132-L137)

The toast is attached to `initEngineGame`'s rejection, but that promise is rejected by **any**
`terminate()` during the load window, not only by a load failure. `terminate` is also the
`'game-concluded'` and `'game-unloaded'` handler
([enginegame.ts:54-55](src/client/scripts/esm/game/misc/enginegame.ts#L54-L55)), and rejects with its
default `'Engine game ended during initialization.'`

Concretely: start a computer game, then hit Abort or Resign before the worker reports `readyok`
(seconds on a cold cache, and reachable on a warm one). The server's `gameconclusion` arrives →
`gameslot.concludeGame()` → `'game-concluded'` → `terminate()` → rejection → the player is told
*"The engine failed to load and resigned the game"* for a game they just ended themselves.
Navigating away mid-load flashes the same message.

**Do this:** move the toast to the point of failure rather than the point of rejection. The three
load-failure sites — the `!window.Worker` branch (lines 94-100), the `initerror` branch (lines
114-118) and `engineWorker.onerror` (lines 120-123) — each already do the same `resignFailedEngine()`
+ reject/terminate pair, so collapse them onto one local helper that also shows the toast, and
reduce the `.catch` in `onlinegame.ts` back to `console.error`. Do **not** put the toast inside
`resignFailedEngine`: its fourth caller is the illegal-engine-move path
([enginegame.ts:274](src/client/scripts/esm/game/misc/enginegame.ts#L274)), which shows its own.

---

## 3. `LIVE_GAME_PERSISTENCE.md` still describes a two-table schema

**File:** [docs/systems/LIVE_GAME_PERSISTENCE.md](docs/systems/LIVE_GAME_PERSISTENCE.md)

The heading became "Three Tables" but the surrounding prose and the matrix did not follow:

- **Line 9** — "live state is split across two tables", directly under the "Three Tables" heading.
- **Line 11-13** — a blank line between the `live_player_games` and `live_engine_games` bullets
  splits one list into two.
- **Line 44** — "Per-player `time_remaining_ms` lives in `live_player_games`." An engine's lives in
  `live_engine_games`.
- **Line 111** — the event matrix's third column is still headed ``live_player_games` Columns
  Updated`, but its rows now describe writes to both participant tables. Retitle it to cover
  participants generally.
- **Line 118** — "cascades to `live_player_games`" on **Draw accepted**; the delete cascades to
  `live_engine_games` too.
- **Line 127** — **Both-disconnected timeout** is annotated "(draw by abandonment / abort)". Engine
  games now conclude as a win for the engine by `disconnect`
  ([gamemanager.ts:595-602](src/server/game/gamemanager/gamemanager.ts#L595-L602)).

---

## 4. The engine page-info mapping exists twice

**Files:** [gamePageController.ts:91-96](src/server/controllers/gamePageController.ts#L91-L96),
[deadgamestate.ts:87-92](src/server/game/gamemanager/deadgamestate.ts#L87-L92)

`produceStaticGameState` declares `engineGame?: EngineGamePageInfo` in its return type but only ever
populates it on the dead branch — it returns `produceDeadStaticGameState(id)` verbatim. The live
branch leaves it undefined, so the controller has to build the same `{ engine, strengthLevel }`
object itself and pick between the two with a ternary.

**Do this:** populate `engineGame` on `produceStaticGameState`'s live branch from
`game.match.engineParticipant`, and reduce the controller to `const { state, game, engineGame,
ratingChanges, moveCount } = resolved;`. One mapping, and `produceStaticGameState`'s declared return
type stops lying about the live path.

While there, `produceDeadStaticGameState`'s JSDoc
([deadgamestate.ts:56-60](src/server/game/gamemanager/deadgamestate.ts#L56-L60)) still enumerates its
return as "its ply count and per-player rating deltas" — it now also returns the engine participant's
page info. Add it.

---

## 5. A load-bearing comment was dropped from `updateLeaderboardsInTransaction`

**File:** [gamelogger.ts:117](src/server/game/gamemanager/gamelogger.ts#L117)

```ts
const leaderboard_id = VariantLeaderboards[match.variant]!;
```

On `redesign` this line carries `// Will always be defined if the game is rated.`, which is what
justifies the non-null assertion. It was removed while the custom-variant branch was being added and
not restored when that branch was reverted.

**Do this:** put the comment back.

---

## 6. The engine schema's time-control refinement sits at the wrong level

**File:** [types.ts:360-371](src/shared/types.ts#L360-L371)

`CreateEngineGameBodySchema` attaches `.refine((body) => clockutil.isTimedControlValid(body.timeControl))`
to the whole object. Its sibling `createseekschem` expresses the identical rule on the field itself —
`time: TimeControlSchema.refine((c) => clockutil.isTimedControlValid(c), { error: 'Invalid clock value.' })`
([createseek.ts:64-66](src/server/game/seeksmanager/createseek.ts#L64-L66)) — and reserves
object-level refinements for genuinely cross-field rules. Field-level also attaches the error to the
right path in the issue it emits.

**Do this:** move the refinement onto `timeControl: TimeControlSchema.refine(...)` and drop the
object-level `.refine`, leaving `CreateEngineGameBodySchema` a plain `z.strictObject`.

PR #1146, opened 2026-07-17
Fix dev-server startup/restart race with the server build

FirePlank:
## Summary
- The unbundled server build (esbuild, `bundle: false`) writes each output file individually, so `wait-on dist/server/server.js` (and nodemon's directory watch on rebuild) could fire before all of a build's files existed — crashing the server with `ERR_MODULE_NOT_FOUND` on startup or restart.
- The client build's `dist/manifest.json` (read eagerly by Nunjucks at server startup) could also still be missing at that point, causing a separate crash.
- Adds a build-complete sentinel file (`dist/server/.build-complete`), rewritten only in esbuild's `onEnd` on a successful build (which only fires once every output file is on disk), and gates both the initial `wait-on` and nodemon's restart watch on that sentinel plus the manifest.

## Test plan
- [x] `npm run type-check` — no new errors introduced (pre-existing unrelated errors confirmed present without this change too)
- [x] `npm run lint` — clean on all touched files
- [x] Ran `npm run dev` from a clean `dist/` — server now starts only after both the server build and client manifest are fully written, no crash

### DIFF
--- a/build/plugins.ts
+++ b/build/plugins.ts
@@ -6,6 +6,9 @@
 
 import type { Plugin, PluginBuild } from 'esbuild';
 
+import fs from 'node:fs';
+import path from 'node:path';
+
 /** Returns an esbuild plugin that logs whenever a build finishes/fails. */
 export function getESBuildLogStatusLogger(successMessage: string, failureMessage: string): Plugin {
 	return {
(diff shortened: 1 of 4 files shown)


Naviary2:
How severe was your issue? What percentage roughly of all startups were you getting a crash, and what was the crash exactly? I need details.

Naviary2:
I need three things:

1. The rough percentage of server startups you encounter this error.
2. Whether the error message is always the exact same, or if it differs.
3. The exact error message for each *type* of error (only one is needed if they are all `ERR_MODULE_NOT_FOUND`), stack trace and all. Sensitive file paths may be redacted.

